Issue #5, opened 2024-11-05
Setup server


StefanNemeth:
- Include Postgres connectivity and docker compose file

PR #10, opened 2024-11-10
Initial Spring boot server setup

StefanNemeth:
Set ups an application server using Gradle for build management, the Sprint Boot Framework, [Lombok](https://github.com/projectlombok/lombok) for avoiding boilerplate code and a PostgreSQL Driver.

It documents the setup process and includes a Docker compose file for setting up the database.

Solves https://github.com/ls1intum/Helios/issues/5


server/application-server/build.gradle
```diff
@@ -0,0 +1,33 @@
...
+	id 'io.spring.dependency-management' version '1.1.6'
+	id 'io.freefair.lombok' version '8.10.2'
+}
+
+group = 'de.tum.cit.aet'
+version = '0.0.1-SNAPSHOT'
+
+java {
+	toolchain {
+		languageVersion = JavaLanguageVersion.of(22)
```
egekocabas, line 14: I think we should use Java 21 since it is the latest LTS

egekocabas: Lets talk this later 👍🏻

egekocabas (approved):

PR #28, opened 2024-11-14
chore(client): setup tailwind and tabler and dockerize client

thielpa:
Setup configs for tailwind and tabler icons.
Also include the client into the docker compose (for development)


TurkerKoc (approved):
LGTM!

PR #29, opened 2024-11-14
chore: Setup notification service and change Java to LTS 21

TurkerKoc:
Basic setup of notification service with spring boot and NATS for communication.

Stream for notification is created in webhook-listener as "notification.>"

Changed Java Version to 21 which is LTS


egekocabas (approved):
🚀

PR #50, opened 2024-11-21
feat(client): Add navigation bar and favicon

thielpa:


egekocabas (approved):
Tried locally and everything works smoothly.
I also really like the layout & icon ☀

PR #52, opened 2024-11-21
chore(deps): Update primeng to v19.0.1

thielpa:
:warning: The client validation does not seem to work right now, this is why the action is failing right now.


egekocabas:
It is really valueable that you add those referances, thanks 🙏🏻

https://github.com/primefaces/primeng/issues/16827#issuecomment-2498068507

egekocabas:
[fyi](https://github.com/primefaces/primeng/issues/16827#issuecomment-2511150789)

> Finally, here’s the exciting news: we’re planning to release 18.Final this week, and 19-rc.1 a week after that. Thank you for your understanding! Always bet on Prime!

thielpa:
Sure, already saw the update and did a quick test if it works, but it seems to need some more migration of some components we use. I will try it next week!

gbanu (approved):

PR #53, opened 2024-11-24
feat: Restructure packages, add workflow run sync + webhook handler

StefanNemeth:
### Description
- Use Lombok annotation `@Log4j2` to avoid boilerplate code
- Get rid of Git/Github sub-packaging, we plan to use Github only
- Re-structure packages using [package by feature convention](http://www.javapractices.com/topic/TopicAction.do?Id=205)
- Use char array within NATS token config as string is deprecated
- Fix bug that didn't pass the proper auth token to nats using docker compose
- Rename RepositoryRepository to GitRepoRepository for clarity reasons
- Implement workflow run entity, handler for workflow run webhook and sync service

Solves #22


egekocabas (approved):
Tested and works smoothly 👍🏻 

We just need to careful about the `WorkflowRun`s since it might be started with different kind of things such as scheduled, dispatch, etc and they might not be connected with a pull request. The `conclusion` field is valuable to determine if the run completed or not.

PR #68, opened 2024-12-01
feat: Connect PR data in client

gbanu:
PR table view

### DIFF
--- a/.github/workflows/build_docker.yml
+++ b/.github/workflows/build_docker.yml
@@ -90,6 +90,7 @@ jobs:
             type=raw,value=latest,enable={{is_default_branch}}
             type=ref,event=branch
             type=ref,event=pr
+            type=sha
 
       - name: Build and push Docker Image
         uses: docker/build-push-action@v6


server/application-server/src/main/java/de/tum/cit/aet/helios/config/SecurityConfig.java
```diff
@@ -0,0 +1,42 @@
...
+import org.springframework.context.annotation.Configuration;
+import org.springframework.security.config.annotation.web.builders.HttpSecurity;
+import org.springframework.security.config.annotation.web.configuration.EnableWebSecurity;
+import org.springframework.security.web.SecurityFilterChain;
+import org.springframework.web.servlet.config.annotation.CorsRegistry;
+import org.springframework.web.servlet.config.annotation.WebMvcConfigurer;
+
+@Configuration
+@EnableWebSecurity
+public class SecurityConfig {
```
gbanu, line 15: @egekocabas I added some simple security config because I was getting "No 'Access-Control-Allow-Origin' header is present on the requested resource." error

client/angular.json
```diff
@@ -45,13 +45,13 @@
...
-                  "maximumError": "1MB"
+                  "maximumWarning": "2mb",
+                  "maximumError": "2.5mb"
                 },
                 {
                   "type": "anyComponentStyle",
-                  "maximumWarning": "4kB",
-                  "maximumError": "8kB"
+                  "maximumWarning": "6kb",
+                  "maximumError": "10kb"
```
egekocabas, line 54: I increased those limits but we need to have a Github action (in the future) to check if the build passes or not. Because when I was trying, the build step was failing due to the limits.

client/src/app/components/pull-request-table/pull-request-table.component.ts
```diff
@@ -0,0 +1,67 @@
...
+import { IconsModule } from 'icons.module';
+import { PullRequestControllerService, PullRequestInfoDTO } from '@app/core/modules/openapi';
+
+
+@Component({
+  selector: 'app-pull-request-table',
+  imports: [CommonModule, TableModule, AvatarModule, TagModule, IconsModule],
+  templateUrl: './pull-request-table.component.html',
+  styles: [`
+    :host ::ng-deep {
```
egekocabas, line 17: I am not an expert on this but is it ok to use `ng-deep` in component or in general?

client/src/app/components/pull-request-table/pull-request-table.component.html
egekocabas, line 1: IMO this loading doesn't look that good, but I think it is the default one from the library 😞 

Dec-02-2024 07-37-49

For the next week maybe we can also talk about loading animation for pages, I am not sure

client/src/app/components/pull-request-table/pull-request-table.component.ts
egekocabas, line 1: In the initial run it creates the table but after switching tabs, the table stuck at loading state. 
(also checked the network tab and it doesn't make a new request)

Dec-02-2024 07-43-06

egekocabas (approved):
Thanks for the client-server connection and also the PR view 🚀 

I left some comments but we really don't need to do them within this PR

PR #86, opened 2024-12-16
feat: Add filter for repositories

thielpa:
This PR adds a global filter, so that only data from the selected repository is fetched from the database.
The repository is selected via the header `X-Repository-Id` with requests to the application server.

:warning: To be sure that all the services have this filter active, please add the `@Transactional` decorator to newly created service classes.


server/application-server/src/main/java/de/tum/cit/aet/helios/filters/RepositoryContext.java
```diff
@@ -0,0 +1,27 @@
...
+
+public class RepositoryContext {
+
+  private RepositoryContext() {}
+
+  private static final InheritableThreadLocal<String> currentRepository =
+      new InheritableThreadLocal<>();
+
+  public static void setRepositoryId(String repositoryId) {
+    System.out.println("Setting tenantId to " + repositoryId);
```
TurkerKoc, line 11: Can we remove this print :)

server/application-server/src/main/java/de/tum/cit/aet/helios/filters/RepositoryFilterAspect.java
```diff
@@ -0,0 +1,27 @@
...
+import org.aspectj.lang.annotation.Pointcut;
+import org.hibernate.Filter;
+import org.hibernate.Session;
+
+@Aspect
+public class RepositoryFilterAspect {
+  @Pointcut(
+      "execution (* org.hibernate.internal.SessionFactoryImpl.SessionBuilderImpl.openSession(..))")
+  public void openSession() {
+    System.out.println("openSession");
```
TurkerKoc, line 14: Can we remove this print as well!

PR #88, opened 2024-12-17
feat: Deployment within PR/Branch detail page

StefanNemeth:
- This code adds a proper details page for both pull requests and branches.
- Enables a user to press on "Deploy" for an unlocked environment, which dispatches a hard-coded branch "main" with the hard-coded "deploy.yml" workflow. So right now it only shows proper output for the main-branch.
- Improved some details such as loading/empty states, environment list search, and a lot more
- Removed the environment commit info placeholder for now until we can show some real data
- Added versioning to environment entity, so we don't run into any race condition for locking at the same time

Notes:
- This adds some new columns where we expect non-null values. So until we have some migration tool set up, we will have to reset our database when we add these kind of columns. Otherwise, the server will crash.
- The way the current state of deployments (in progress, successful) are handled right now is very hacky. When dispatching a workflow, it takes some time until a deployment object is created by GitHub. So right now, we directly set a "deploying" column on the environment and then update that when we get a new deployment object. So it does not handle any failures, etc.. We could watch the workflow instead of the deployment, that's open for discussion.
- There will be some additional questions, like: Are we able to unlock an environment if there is currently running a deployment workflow for it?

## Screenshots

image

image

image


client/src/app/components/environments/environment-list/environment-list-view.component.ts
TurkerKoc, line 1: - That’s search functional for environments is great! 
Another idea for searching: Since there are typically not that much of environments for a repo, we can maybe search installed apps instead of environments on search. So that developers can see which environments they can deploy for their app.

client/src/app/components/environments/environment-list/environment-list-view.component.html
```diff
@@ -0,0 +1,75 @@
+<div class="flex items-center justify-between mb-3">
+  <input pInputText id="commit-hash" (input)="onSearch($event)" [value]="searchInput()" type="text" placeholder="Search for installed systems" class="w-1/3" />
+  @if (!hideLinkToList()) {
+    <a [routerLink]="'/environments'" class="p-button p-2 p-button-secondary self-end">Manage environments</a>
```
TurkerKoc, line 4: Looks like ‘Manage Environments’ link is forwarded directly to /environments, can you change it to /repo/:repoId/environment/list

client/src/app/components/environments/environment-list/environment-list-view.component.html
```diff
@@ -0,0 +1,75 @@
...
+            @if (environment.serverUrl) {
+              <a [href]="getFullUrl(environment.serverUrl)" target="_blank" class="p-button p-button-secondary p-2" (click)="$event.stopPropagation()"
+                ><i-tabler name="external-link" class="mr-1" />Open</a
+              >
+            }
+            <button (click)="unlockEnvironment.mutate(environment); $event.stopPropagation()" class="p-button p-button-danger p-2">
+              <i-tabler name="lock-open" class="mr-1" />Unlock
+            </button>
+          }
+
```
TurkerKoc, line 49: Looks like you’ve removed the ‘Open link’ button from environment list view. Can we add it back. Since we are currently allowing users to configure URL for each environments on edit.

client/src/app/app.routes.ts
TurkerKoc, line 1: After login it forwards to /project/:id, this should be changed. I think @gbanu will take care of it.

TurkerKoc (approved):
I've tested the deployment in a dummy repo. It works like a charm. Thanks! Other than that I've added some comments which can be implemented later on. Good Job!

Issue #113, opened 2024-12-22
Connect Project Settings with deployment


TurkerKoc:
- Fix labeling workflows in project settings (one build and one deployment workflow should be chosen)
- Triggering deployment should dispatch the chosen deployment yaml.
- refactor project setting choosing only one deployment and build yml should be allowed

PR #125, opened 2024-12-27
feat: Deploying the chosen workflow in Project Settings

TurkerKoc:
Solves #113


client/src/app/pages/project-settings/project-settings.component.ts
```diff
@@ -180,15 +185,32 @@ export class ProjectSettingsComponent {
...
+    });
   }
 
   onChangeLabel(workflow: WorkflowDto) {
-    this.workflowLabelMutation.mutate({ path: { workflowId: workflow.id }, body: workflow.label });
+    const label = workflow.label;
+    if (label === 'DEPLOYMENT' || label === 'BUILD') {
+      const existingLabel = this.workflows().find(wf => wf.label === label && wf.id !== workflow.id);
+      if (existingLabel) {
+        console.warn(`Only one workflow can be labeled as ${label}.`);
```
thielpa, line 209: question(ux): maybe we should add a confirmation modal here to signal to the user, that the existing label is changed. There should be a component from primeng

thielpa (changes requested):
Nice one, thanks :100: ! Please change the best-practice stuff, apart from that it's ready to merge from my point of view :) If the modal stuff should be done in a separate PR, please open an issue for that

thielpa (approved):

PR #127, opened 2024-12-27
docs: Improve documentation on setup and running the app

thielpa:


TurkerKoc (approved):
That looks great! Thanks for the efforts 🚀. I think we need to add 1-2 minor documentation such as Architecture, prod setup guide, however we can finalize it in [another task](https://github.com/ls1intum/Helios/issues/187).

PR #136, opened 2024-12-28
feat: Improve pull request table view and add reviewer and labels

thielpa:
image

Currently this uses only mocked labels and workflow status, we have to add the labels server-side first (#135)


TurkerKoc (approved):
LGTM! Great work! 

One minor thing that we can deal later on: Github external link button looks little bit odd on bigger screens.

image

PR #137, opened 2024-12-28
refactor: Page layout and include repository data

thielpa:


client/src/app/pages/project-settings/project-settings.component.html
```diff
@@ -1,127 +1,137 @@
...
-                  <button pButton label="Delete Group" class="p-button-danger p-button-sm self" (click)="deleteGroup(group)"></button>
-                }
+              <!-- Horizontal line between panels -->
+              @if (group !== localGroupedWorkflowsArray()[localGroupedWorkflowsArray().length - 1]) {
+                <div class="w-10 h-px bg-gray-300 mt-7"></div>
+              }
+            </div>
+          }
+        </div>
+      } @else {
```
TurkerKoc, line 54: With this change, there should be no groups in order to see this 'No groups available' dialog with + button to add a new group. Once we add a group, we are no longer able to add more groups. We should put Add Group button back

client/src/app/pages/project-settings/project-settings.component.html
```diff
@@ -1,127 +1,137 @@
...
-    @if (!isPending() && !isError()) {
-      <div class="space-x-2">
-        <button pButton label="Add New Group" class="p-button-sm" (click)="showAddGroupDialog = true"></button>
-        <button pButton label="Update Groups" class="p-button-success p-button-sm" (click)="updateGroups()" [disabled]="isPending()"></button>
-      </div>
+<div>
+  <app-page-heading [repositoryId]="repositoryId()">
+    <div class="flex justify-between items-center w-full" heading>
+      <div>Repository Settings</div>
+      <!-- TODO: Remove and replace by update handlers directly on dropdowns -->
```
TurkerKoc, line 5: We can put Add Group button back here

TurkerKoc (changes requested):
Thanks for the great job! New layout with headers matching looks really nice! There is one minor issue in Project Settings and some conflicts with the new project settings page as well. It's good to go we update project settings! 🚀

client/src/app/pages/project-settings/project-settings.component.html
thielpa: I put the button here, i think this is more intuitive for the user. Otherwise he/she has to search on the page. Hope this is fine, otherwise I can change it again.
image

TurkerKoc (approved):

PR #142, opened 2025-01-02
feat: Improve branches table view

thielpa:
image


TurkerKoc (changes requested):
UI Looks perfect! Thanks for the great efforts!! 

One issue I found after testing: 

- When I create a branch and push it in the repository being used in Helios, once webhook comes I get this error
image

- Then when the branches page refreshed everything fails with an error 'user' is null
image

- And branches view never loads:
image


There is something wrong with parsing users.

TurkerKoc (approved):
LGTM!

PR #161, opened 2025-01-06
feat: Check user permissions for repository

gbanu:
PR #147 and PR #148 combined
using token from the .env to dispatch a workflow

### DIFF
--- a/client/src/app/components/deployment-selection/deployment-selection.component.ts
+++ b/client/src/app/components/deployment-selection/deployment-selection.component.ts
@@ -32,6 +32,7 @@ export class DeploymentSelectionComponent {
 
   handleDeploy = (environment: EnvironmentDto) => {
     this.currentEnvironmentId = environment.id;
-    this.deployEnvironment.mutate({ body: { environmentId: environment.id, branchName: this.sourceRef() } });
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    this.deployEnvironment.mutate({ body: { environmentId: environment.id, branchName: this.sourceRef() } } as any);
   };
 }
--- a/client/src/app/components/deployment-selection/deployment-selection.component.ts
+++ b/client/src/app/components/deployment-selection/deployment-selection.component.ts
@@ -1,9 +1,10 @@
-import { Component, inject, input } from '@angular/core';
-import { EnvironmentListViewComponent } from '../environments/environment-list/environment-list-view.component';
-import { injectMutation, injectQueryClient } from '@tanstack/angular-query-experimental';
-import { deployToEnvironmentMutation, getAllEnvironmentsQueryKey, getEnvironmentByIdQueryKey } from '@app/core/modules/openapi/@tanstack/angular-query-experimental.gen';
+/* eslint-disable prettier/prettier */
+import { Component, computed, inject, input } from '@angular/core';
 import { EnvironmentDto } from '@app/core/modules/openapi';
+import { deployToEnvironmentMutation, getAllEnvironmentsQueryKey, getEnvironmentByIdQueryKey, getRepoPermissionsOptions } from '@app/core/modules/openapi/@tanstack/angular-query-experimental.gen';
+import { injectMutation, injectQuery, injectQueryClient } from '@tanstack/angular-query-experimental';
 import { MessageService } from 'primeng/api';
+import { EnvironmentListViewComponent } from '../environments/environment-list/environment-list-view.component';
 
 @Component({
   selector: 'app-deployment-selection',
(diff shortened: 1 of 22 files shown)


server/application-server/src/main/java/de/tum/cit/aet/helios/permission/UserPermissionController.java
```diff
@@ -0,0 +1,26 @@
...
+public class UserPermissionController {
+  private GitHubService gitHubService;
+
+  public UserPermissionController(GitHubService gitHubService) {
+    this.gitHubService = gitHubService;
+  }
+
+  @GetMapping("/repository/{repoId}")
+  public ResponseEntity<GHPermissionType> getRepoPermissions(
+      @RequestHeader("Authorization") String token, @PathVariable Long repoId) {
```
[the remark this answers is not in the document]
thielpa: issue: due to the types conflicts in the client when setting the header to required, we should handle the Auth token like the Repository-Id header (add interceptor like in RepositoryInterceptor and create context. 
@TurkerKoc Maybe there is also another way with the JwtAuthenticationToken stuff you implemented to retrieve the token from some context of spring boot, but not an expert on this

server/application-server/src/main/java/de/tum/cit/aet/helios/github/GitHubService.java
```diff
@@ -206,4 +221,59 @@ public Iterator<GitHubDeploymentDto> getDeploymentIterator(
...
+      return responseBody.split("&")[0].split("=")[1];
+    } catch (IOException e) {
+      log.error("Error occurred while exchanging GitHub token: {}", e.getMessage());
+      throw new RuntimeException("GitHub token exchange failed", e);
+    }
+  }
+
+  private String getUsernameFromToken(String bearerToken) throws IOException {
+
+    String token = bearerToken.replace("Bearer", "").trim();
```
thielpa, line 266: question: afaik @TurkerKoc used `org.springframework.security.oauth2.jwt.Jwt` for jwt decoding in `KeycloakJwtAuthenticationConverter.java`. Is this also possible here?

client/src/app/components/deployment-selection/deployment-selection.component.ts
```diff
@@ -16,13 +17,26 @@ export class DeploymentSelectionComponent {
   queryClient = injectQueryClient();
 
   sourceRef = input.required<string>();
+  repositoryId = input.required<number>();
+
+  permissionsQuery = injectQuery(() => ({
+    ...getRepoPermissionsOptions({
+      path: { repoId: this.repositoryId() }
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
thielpa, line 25: nitpick: this is pretty dirty. I would recommend to change the server code as described in my comment below

client/src/app/components/deployment-selection/deployment-selection.component.ts
```diff
@@ -16,13 +17,26 @@ export class DeploymentSelectionComponent {
...
+  hasDeployPermissions = computed<boolean>(() => this.permissionsQuery.data() === 'WRITE' || this.permissionsQuery.data() === 'ADMIN');
+  hasAdminPermission = computed<boolean>(() => this.permissionsQuery.data() === 'ADMIN');
 
   private currentEnvironmentId: number | null = null;
 
   deployEnvironment = injectMutation(() => ({
     ...deployToEnvironmentMutation(),
     onSuccess: () => {
-      this.queryClient.invalidateQueries({ queryKey: getAllEnvironmentsQueryKey() });
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
thielpa, line 38: nitpick: same here

client/src/app/components/deployment-selection/deployment-selection.component.ts
```diff
@@ -32,6 +46,7 @@ export class DeploymentSelectionComponent {
 
   handleDeploy = (environment: EnvironmentDto) => {
     this.currentEnvironmentId = environment.id;
-    this.deployEnvironment.mutate({ body: { environmentId: environment.id, branchName: this.sourceRef() } });
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
thielpa, line 49: nitpick: same here

client/src/app/components/environments/environment-list/environment-list-view.component.ts
```diff
@@ -23,6 +23,8 @@ export class EnvironmentListViewComponent {
 
   editable = input<boolean | undefined>();
   deployable = input<boolean | undefined>();
+  hasUnlockPermissions = input<boolean | undefined>();
```
thielpa, line 26: tip: you don't have to specify undefined here, it is automatically added to the type when you don't add a default value or use input.required()

server/application-server/src/main/java/de/tum/cit/aet/helios/deployment/DeploymentController.java
```diff
@@ -50,13 +51,13 @@ public ResponseEntity<DeploymentDto> getLatestDeploymentByEnvironmentId(
   }
 
   @PostMapping("/deploy")
-  public ResponseEntity<String> deployToEnvironment(@RequestBody DeployRequest deployRequest) {
+  public ResponseEntity<String> deployToEnvironment(
+      @RequestHeader("Authorization") String token, @RequestBody DeployRequest deployRequest) {
     try {
-      deploymentService.deployToEnvironment(deployRequest);
+      deploymentService.deployToEnvironment(deployRequest, token);
```
thielpa, line 57: issue: please check if the token is set before calling the deploy function (not sure if this is an authenticated route though) and return an unauthenticated error instead of the internal server error

thielpa (changes requested):
Really nice work with the tokens and auth! Works like a charm on my local machine, but there are some comments we should discuss before merging it

server/application-server/src/main/java/de/tum/cit/aet/helios/github/GitHubService.java
TurkerKoc: again please refer to this [file](https://github.com/ls1intum/Helios/blob/main/server/application-server/src/main/java/de/tum/cit/aet/helios/auth/AuthService.java)

PR #162, opened 2025-01-06
fix: Handle commit users with no linked gh account

thielpa:


egekocabas (approved):
Tested with the Helios & Artemis data and the syncing works fine 👍🏻

PR #179, opened 2025-01-06
fix: Harmonize client, fix table filters and remove settings icon

thielpa:


client/src/app/core/services/keycloak/keycloak.service.ts
```diff
@@ -21,6 +21,19 @@ export class KeycloakService {
...
+  decodedToken = computed(() => {
+    const token = this.keycloak.token;
+    if (!token) {
+      return null;
+    }
+    const [, payload] = token.split('.');
+    return JSON.parse(atob(payload));
+  });
+
+  isCurrentUser(login?: string) {
```
TurkerKoc, line 33: current user is failing to match because of the uppercase letters. Seems like keycloak is storing my username as turkerkoc however it's TurkerKoc in github

egekocabas: I looked it up, and GitHub usernames are not case-sensitive. It shows your name as `TurkerKoc`, but the username is unique as if it’s all lowercase. 

So this implementation shouldn't be a problem i think.

thielpa: Ok, for safety reasons I added the check nonetheless :)

egekocabas (approved):
Since we first merge https://github.com/ls1intum/Helios/pull/182, it introduced conflicts with current PR, but other than that I approve it

PR #199, opened 2025-01-11
feat: Custom Keycloakify theme

gbanu:
copied and adjusted the theme from Hephaestus
Screenshot 2025-01-12 at 18 20 44


egekocabas (changes requested):
Thanks it works perfect 🥳 

But i have couple of requests:
- Can you search across the `keycloakify` folder and find all `hephaestus` words and change with `helios`. (PS: Also for the folder names, such as [this](https://github.com/ls1intum/Helios/tree/keycloakify/keycloakify/dist_keycloak/theme/hephaestus)) 
- Since we only updated the `compose.yaml`, this PR doesn't have production changes right? Can you also find a way to use this in production? (Might need to build & deploy)

egekocabas (approved):
Created `./keycloakify/dist_keycloak` in production and copied jar files & edited prod `realm.json` to use helios theme 👍🏻 

image

Also tested locally one more time and works perfect thank you 🚀

egekocabas:
@gbanu but i don't understand why did you change the prod compose file's webhook listener part? i don't know whether the new version works or not?

Issue #194, opened 2025-01-07
Improve Environment Edit page


thielpa:
In the env. edit page the header is not shown and also in the sidebar the active item is gone

![image](https://github.com/user-attachments/assets/0a5bb566-29d2-4e30-9a69-e51942930187)

Issue #166, opened 2025-01-06
Managers should be able to disable deployments for desired environments


egekocabas:
Since we are getting all the environments defined in GitHub, managers should be able to edit environment and disable that environment to deployments

![Image](https://github.com/user-attachments/assets/8151873a-bd2e-4b89-8711-2245be89230d)

Issue #164, opened 2025-01-06
Order of environments change when deploy button is clicked


egekocabas:
It was previously in the middle but as soon as i click deploy button it went down to the buttom of the list, when i refresh the page it stays at the bottom as well. Also unlocking doesnt change the order (still stays at the bottom). Maybe this is related with server sending in this order i dont know 

![Image](https://github.com/user-attachments/assets/93e1e384-acad-475f-816a-1e92731e97db)

PR #200, opened 2025-01-11
feat: Add functionality to disable environments

egekocabas:
### Motivation
Closes https://github.com/ls1intum/Helios/issues/166, closes https://github.com/ls1intum/Helios/issues/164, closes https://github.com/ls1intum/Helios/issues/194

### Description
- `enabled` field added to `Environment` entity
- environment `update` endpoint refactored to update the `enabled` field correctly
- `environmentdto` contains new `enabled` field 
- new endpoint `/api/environments/enabled` added
  - returns all enabled environments
- if the environment list is **editable** **and user loggedin** (will be replaced by manager/admin check later on), show `Disabled` label near edit button, otherwise do not show a label
- Dynamically either call `getAllEnvironments` or `getAllEnabledEnvironments` in `environment-list-view.component.ts`
- Enabled checkbox added in environment edit page

#### Not Logged-in User
image
#### Logged-in User
image

#### Logged-in User
https://github.com/user-attachments/assets/cc729ea9-c651-4205-bd67-0f7abac255e6

#### Both Logged-in User & Not Logged-in User Screenshots (View is same)
image
image

**Edit 1:**
### Disable environment when it is locked

https://github.com/user-attachments/assets/f63a1af9-49e2-4409-96df-1e20143f9f06

**Edit 2:**
### Header added
image

**Edit 3:**
### Fix active routing for env edit page
image


client/src/app/components/environments/environment-list/environment-list-view.component.ts
```diff
@@ -43,14 +49,21 @@ export class EnvironmentListViewComponent {
 
   searchInput = signal<string>('');
 
-  environmentQuery = injectQuery(() => getAllEnvironmentsOptions());
+  // Dynamically determine the query function & query key
+  // If the component is editable and logged in (assuming manager), we want to show all environments; otherwise only enabled environments
+  // In PR/Branch View, editable is false, so we only show enabled environments
+  // TODO: We need to also check if the user has the correct permissions to view all the environments
```
TurkerKoc, line 55: I think everybody can see all the environments, even if they are not logged in right?

client/src/app/components/forms/environment-edit-form/environment-edit-form.component.ts
```diff
@@ -67,6 +69,7 @@ export class EnvironmentEditFormComponent implements OnInit {
       installedApps: [this.environment()?.installedApps || []],
       description: [this.environment()?.description || ''],
       serverUrl: [this.environment()?.serverUrl || ''],
+      enabled: [this.environment()?.enabled || false],
```
TurkerKoc, line 72: Should we allow managers to disable the environment which is currently locked? 
2 options:
- We won't allow managers to disable currently locked environment. (they can acquire lock and disable it manually)
- We allow them and automatically unlock the environment before disabling it.

TurkerKoc: Currently, environment stays locked and counter is still counting after disabling the environment

server/application-server/src/main/java/de/tum/cit/aet/helios/environment/EnvironmentService.java
```diff
@@ -165,6 +175,7 @@ public Optional<EnvironmentDto> updateEnvironment(Long id, EnvironmentDto enviro
               if (environmentDto.serverUrl() != null) {
                 environment.setServerUrl(environmentDto.serverUrl());
               }
+              environment.setEnabled(environmentDto.enabled());
```
TurkerKoc, line 178: Maybe checking locking status here?

TurkerKoc (changes requested):
Great work! Nice that you've added enabling/disabling environments along with controller to filter to only enabled environments. 

I've added a minor comment related to disabling currently locked environments. Maybe we should address this before merging.

egekocabas:
@TurkerKoc Much thanks for spotting an issue. I choose to not let admins to disable environments if its locked. They need to manually unlock first and then disable it. I choose this approach because the admins might accidentally disable environments and it looks better to me that if they concisely unlock fist. What do you think?

Also added and order for environemnt for this issue https://github.com/ls1intum/Helios/issues/164

egekocabas:
It make sense to also fix https://github.com/ls1intum/Helios/issues/194 issue within this PR, today i will push the changes

egekocabas:
@TurkerKoc ready for review for a second round 🥊  new screenshots are available in description

TurkerKoc (changes requested):
Thanks for making all the changes! Everything looks great. However, I think you've accidentally deleted your changes while merging main into your branch. 

- Order of environments still changing after deployment
image
image

- I'm still able to disable locked environment
image

I saw that you've also included #194 in this PR. If you are planning to close that issue:
- Can we add page header into edit page. 
- No page is selected as current page in the navigation bar. Maybe we can select environments?
image

client/src/app/components/environments/environment-list/environment-list-view.component.ts
```diff
@@ -45,12 +51,19 @@ export class EnvironmentListViewComponent {
...
```
TurkerKoc, line 57: Since we pushed permission changes already. Can we add it here as well?

server/application-server/src/main/java/de/tum/cit/aet/helios/environment/EnvironmentLockHistoryRepository.java
```diff
@@ -1,22 +1,52 @@
...
-      String lockedBy
+  @Query("""
+        SELECT elh
+        FROM EnvironmentLockHistory elh
+        WHERE elh.environment = :environment
+          AND elh.lockedBy = :lockedBy
+          AND elh.unlockedAt IS NULL
+        ORDER BY elh.lockedAt DESC
+      """)
+  Optional<EnvironmentLockHistory> findLatestLockForEnvironmentAndUser(
```
TurkerKoc, line 30: We were sending only one element by using 'findTopBy' tag. However, now if user locks two environments, '_/environments/user_controller_' endpoint fails as following:

image

Maybe locking more than one environment should not be allowed to any user. I will create a new task for this one. Until then can you change this query to return only one lock?

egekocabas:
- ordering of environments fixed
- counter issue fixed 
https://github.com/user-attachments/assets/8e86114f-2c0d-4e05-8877-aafc50ae11de

client/src/app/components/environments/environment-list/environment-list-view.component.ts
egekocabas: Since the permission handling will change in here https://github.com/ls1intum/Helios/issues/207 i propose that we solve/fix this separately

server/application-server/src/main/java/de/tum/cit/aet/helios/environment/EnvironmentLockHistoryRepository.java
egekocabas: Will be fixed in https://github.com/ls1intum/Helios/issues/209

TurkerKoc (approved):
LGTM! Thanks for the efforts 🚀

Issue #106, opened 2024-12-21
Required steps to setup new repo in Helios


egekocabas:
- This is the least priority for this iteration since we have other features we want to finish before the 7th

TurkerKoc:
- We need to create a documentation for how to setup ls1intum repo in Helios.
- Include all the required steps
- Implement necessary updates on Helios

egekocabas:
@thielpa https://github.com/ls1intum/Helios/blob/main/server/application-server/src/main/java/de/tum/cit/aet/helios/github/GitHubClientManager.java#L279-L299 this section might be valuable, this is the place where we get the installation id if it is not present in the `.env` file

PR #203, opened 2025-01-12
feat: Implement custom protection rule flow

egekocabas:
### Motivation
We needed a way to handle GitHub’s `deployment_protection_rule` events so that our application can approve or reject environment deployments based on custom rules—for example, whether the workflow was triggered by our app itself. However, the `org.kohsuke.github` library does not natively support this event type. Additionally, we wanted a robust, centralized mechanism for managing both GitHub App credentials (JWT) and personal access tokens (PATs), ensuring our tokens remain valid and fresh during usage. We also needed to fix a deployment webhook error related to the `WAITING` state, which was previously unhandled by `org.kohsuke.github`.

### Description
- Custom Webhook Listener for Deployment Protection Rule
  - Introduced a new webhook listener specifically for `deployment_protection_rule` events.
  - Since `org.kohsuke.github` didn't implemened the `deployment_protection_rule` event type,  a custom message handler interface is added.
- Dual-Credential `GitHubClientManager`
  - Created `GitHubClientManager`, which handles both `GitHub App credentials` (JWT-based) and `PAT` credentials.
  - This change gives us flexibility. `PAT` can be used for locally and `GitHub App credentials` can be used for the production.
  - Refreshes tokens (`GitHub App`) every 20 minutes to prevent expiration.
- `GitHubFacade` Introduction
  - Implemented `GitHubFacade` (and its `GitHubFacadeImpl`) as a **wrapper** around the raw GitHub API client.
  - Ensures the application always uses the most up-to-date GitHub client with **valid tokens** (through `GitHubClientManager`).
  - Replaced direct usage of `GitHub` beans throughout the codebase with the new facade.
  - Future `GitHub` method calls can be added to the facade interface and its implementation as needed.
- Fix for Deployment event's `WAITING` State
  - Resolved an issue where `org.kohsuke.github.GHDeploymentState` did not implement the `WAITING` state, causing exceptions if the status was `WAITING`.
  - Introduced a check to handle the `WAITING` state gracefully in `Deployment.mapToState`.

#### General
- [x] Documentation is updated.

#### Server


server/application-server/src/main/java/de/tum/cit/aet/helios/github/GitHubClientManager.java
```diff
@@ -0,0 +1,321 @@
...
+   */
+  private volatile String currentToken;
+  /**
+   * The expiration time of the current token.
+   */
+  private volatile Instant tokenExpirationTime;
+
+  private final ReentrantLock lock = new ReentrantLock();
+
+  @Value("${http.cache.ttl}")
```
thielpa, line 87: please set a default value when no env value is specified - if possible. This makes it easier for e2e/integration test setup later on :)

I saw you added defaults in the application.yaml files, but I guess it is better to handle it here globally

server/application-server/src/main/java/de/tum/cit/aet/helios/github/GitHubClientManager.java
```diff
@@ -0,0 +1,321 @@
...
+  }
+
+  /**
+   * Creates a GitHub client with caching enabled.
+   *
+   * @return the GitHub client
+   */
+  public GitHub createGitHubClientWithCache() {
+    if (authType == null) {
+      return GitHub.offline();
```
thielpa, line 201: suggestion: please add an warning log here, otherwise it is hard to debug why the github client is offline (already had this experience when fixing a bug)

server/application-server/src/main/java/de/tum/cit/aet/helios/github/GitHubConfig.java
```diff
@@ -26,85 +17,52 @@ public class GitHubConfig {
   @Value("${github.authToken}")
   private String ghAuthToken;
 
-  @Value("${github.cache.enabled}")
-  private boolean cacheEnabled;
+  @Value("${github.appName}")
```
thielpa, line 20: suggestion: please also add defaults here - if possible

server/application-server/src/main/java/de/tum/cit/aet/helios/github/GitHubFacade.java
```diff
@@ -0,0 +1,28 @@
...
+/**
+ * This facade is used as a wrapper instead of directly using the GitHub bean.
+ * It ensures that the implementation uses the most up-to-date GitHub client
+ * with the latest JWT token and calls the respective GitHub methods.
+ *
+ * <p>If new methods of GitHub need to be called, this interface and implementation
+ * should be updated to add that method. Use the same method signature as in GitHub.
+ * </p>
+ */
+public interface GitHubFacade {
```
thielpa, line 18: praise: :100:  using the GitHubFacade and refreshing the client with this

server/application-server/src/main/java/de/tum/cit/aet/helios/github/HttpClientConfig.java
```diff
@@ -0,0 +1,80 @@
...
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.context.annotation.Bean;
+import org.springframework.core.env.Environment;
+import org.springframework.stereotype.Component;
+
+@Log4j2
+@Component
+public class HttpClientConfig {
+
+  @Value("${http.cache.enabled}")
```
thielpa, line 18: suggestion: defaults

thielpa (changes requested):
Thanks for the PR, codewise this complex topic looks pretty clean and well-thought!

Please just make sure that you add the added environment variables also to the documentation. Otherwise I just found minor suggestions :)

:warning: Looked into it only codewise yet.

server/application-server/src/main/java/de/tum/cit/aet/helios/deployment/Deployment.java
```diff
@@ -83,4 +98,38 @@ public static State mapToState(GHDeploymentState ghState) {
...
+      String toStringValue = ghDeploymentStatus.toString();
+      // Use a regex to extract the value of the `state` field
+      Pattern pattern = Pattern.compile("state=([a-zA-Z_]+)(?:,|\\])");
+      Matcher matcher = pattern.matcher(toStringValue);
+      if (matcher.find()) {
+        return matcher.group(1).toUpperCase();
+      } else {
+        return "";
+      }
+    } catch (Exception e) {
```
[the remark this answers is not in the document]
egekocabas: While working with the `org.kohsuke.github` library, I noticed that the method signatures can sometimes be misleading, as they do not always explicitly indicate that certain methods may throw exceptions. 

So just to be in the safe spot I wrapped the entire logic in a `try/catch` so that for both `ghDeploymentStatus.toString()` or `patter matching` might throw some exception, i wanted to cover all of them. 

I updated the code to explicitly catch the `PatternSyntaxException` thrown by the `Pattern.compile` and i also included generic Exception as fallback. 

I also added logging to identify issues better 👍🏻

docs/development/local/setup.rst
TurkerKoc, line 1: I think we should also add the documentation related to adding protection rule to the existing environments. Other than that documentation was simple and straightforward

docs/development/local/setup.rst
```diff
@@ -56,31 +56,287 @@ You can then change/set the environment variables in the `.env` file to your nee
...
+---
+
+**Other Required Variables** (server `.env`):
+
+- ``DATASOURCE_URL``: URL to the database
+- ``DATASOURCE_USERNAME``: Database username
+- ``DATASOURCE_PASSWORD``: Database password
+- ``NATS_SERVER``: NATS server URL
+- ``NATS_AUTH_TOKEN``: Authorization token for NATS server. This token is used to authenticate different services with the NATS server.
+- ``REPOSITORY_NAME``: Name of the repository that should be synced (e.g. `ls1intum/Helios`)
```
thielpa, line 118: question: is this still needed? Don't we have already all the information from the GH app / PAT? For me this looks like duplicated information. If this can be removed, we can add it as a ticket to the next sprint.

docs/development/local/setup.rst
```diff
@@ -56,31 +56,287 @@ You can then change/set the environment variables in the `.env` file to your nee
...
+
+- ``GITHUB_AUTH_TOKEN``: Your personal access token (PAT). *(This enables basic GitHub operations but not custom deployment protection rules.)*
+
+**If using a GitHub App** (recommended for full functionality):
+
+- ``GITHUB_APP_NAME``: Name of your GitHub App. After creation, you can confirm the "actual" URL-safe name in the GitHub UI or from the App URL in your browser. For example, if the GitHub UI shows "Helios (AET)" but the URL is ``https://github.com/organizations/ls1intum/settings/apps/helios-aet``, then ``GITHUB_APP_NAME=helios-aet``.
+- ``GITHUB_APP_ID``: Numeric ID of your GitHub App (from GitHub App settings).
+- ``GITHUB_CLIENT_ID``: Client ID (from GitHub App settings).
+- ``GITHUB_PRIVATE_KEY_PATH``: **Absolute** or **relative** path to the **PKCS#8**-formatted private key file associated with your GitHub App. (See note below on converting from PKCS#1 to PKCS#8.)
+- ``ORGANIZATION_NAME``: Name of the GitHub organization/user **only if** you do **not** know the installation ID and want Helios to **auto-detect** it at runtime. Helios will retrieve the GitHub App installations for the given organization/user name and pick the correct installation.
```
thielpa, line 97: For me it is not working without setting the organization_name variable. Maybe there is something wrong with the defaults?

thielpa (commented):
Tested it out, worked for me following the docs :) Thanks for the effort!

TurkerKoc (approved):
I looked at the code and tried client manager with and without Github App. Everything works fine! Thanks for the efforts

egekocabas:
> Tested it out, worked for me following the docs :) Thanks for the effort!

@thielpa Thank you so much for testing 💯, we will merge this one right now but will fix those points in couple days in another PR 👍🏻

docs/development/local/setup.rst
egekocabas: leaving this as it is for https://github.com/ls1intum/Helios/pull/273 since the repo sync mechanism will be implemented in https://github.com/ls1intum/Helios/issues/106

egekocabas:
@thielpa You review comments is will be fixed in https://github.com/ls1intum/Helios/pull/273 🙏🏻

PR #211, opened 2025-01-14
fix: All Error messages

TurkerKoc:
- Now we have standardized error handling.
- Created ApiError class.
- Created GlobalExceptionHandler which is catching exceptions and creating ApiError entity and returning it to client.
- Please add your exception handler if you create a new exception type or if it's not already handled in GlobalExceptionHandler.
- Changed 401 error message:
Screenshot 2025-01-14 at 02 00 20


server/application-server/src/main/java/de/tum/cit/aet/helios/error/GlobalExceptionHandler.java
```diff
@@ -0,0 +1,93 @@
...
+
+  // -- 500 INTERNAL SERVER ERROR (FALLBACK) -------------
+  @ExceptionHandler({Exception.class, IOException.class})
+  public ResponseEntity<ApiError> handleGeneralException(
+      Exception ex, HttpServletRequest request) {
+
+    ApiError error = new ApiError();
+    error.setStatus(HttpStatus.INTERNAL_SERVER_ERROR.value());
+    error.setError(HttpStatus.INTERNAL_SERVER_ERROR.getReasonPhrase());
+    error.setMessage("Error: " + ex.getMessage());
```
thielpa, line 73: issue: not sure about sending the exception message to the client and show it there as it might leak sensitive information

thielpa (changes requested):

thielpa (approved):
LGTM now :)

Issue #169, opened 2025-01-06
If lock or unlock is clicked, it doesn't immidiately effect the sidebar counter


egekocabas:
Below is one example but it is true for the other case as well

https://github.com/user-attachments/assets/102c2065-9746-4228-b909-05e99d5e21f8

PR #217, opened 2025-01-18
fix: Use `invalidateQueries` for immediate lock timer synchronization

egekocabas:
Refer to https://github.com/ls1intum/Helios/issues/169

### Motivation
Synchronize updates between `main-layout` and `environment-list-view` components to eliminate delays in lock timer synchronization.

### Description
Introduced `EnvironmentStateService` to manage environment update events. The `environment-list-view` component emits update events upon **unlocking environments**, and the `deployment-selection` component emits update events upon **deploying environments**. The `main-layout` component subscribes to these events to refetch environment data instantly.

#### Branch (Deploy - Unlock)
https://github.com/user-attachments/assets/81deb17e-c213-4cf7-8056-9e9c298f8050
#### PR (Deploy - Unlock)
https://github.com/user-attachments/assets/cb54ba78-f496-436e-9991-8df2efd9129b
#### Environment list (Unlock)
https://github.com/user-attachments/assets/2e385e1c-3421-4c33-8694-a83642c69812


client/src/app/pages/main-layout/main-layout.component.ts
```diff
@@ -96,6 +101,16 @@ export class MainLayoutComponent implements OnInit, OnDestroy {
   });
 
   ngOnInit() {
+    // Subscribe to environment update events
+    this.subscription = new Subscription();
```
thielpa, line 105: question: why do we need the environmentStateService? Can't we invalidate just all locks in above success handler? I don't quite get what information we miss there :)

thielpa (approved):
Nice work, just having one question about the approach. Thanks!

client/src/app/pages/main-layout/main-layout.component.ts
egekocabas: You are so right 😅 I don't know why initially I didn't use `invalidateQueries`...

egekocabas: I removed the observable approach and use `invalidateQueries`. I tested and it is working as before 👍🏻

egekocabas: As a validation, i will request review again and then we can merge it right away

thielpa (approved):
Nice work! Maybe adjust the title of the PR now, it does not seem to match anymore :)

Issue #163, opened 2025-01-06
Workflow Grouping Doesn't Fully Shown in PR/Branch View


egekocabas:
Below is settings page:

![Image](https://github.com/user-attachments/assets/db971731-0c0f-4b7f-8be2-c2bc61d853c7)

Below is PR view:

![Image](https://github.com/user-attachments/assets/7a6138c6-a8db-4d2e-9e39-3989e79d4670)

I think its because in the PR/Branch view, the container doesn't fit to the entire screen.

![Image](https://github.com/user-attachments/assets/76445ed7-ea08-484d-bad7-34cb4a14a9b0)

egekocabas:
It looks like there is no problem 😅 

![Image](https://github.com/user-attachments/assets/fc15017c-3b98-44c2-88a7-28dc8dc004a8)

![Image](https://github.com/user-attachments/assets/be0f669b-e5dc-47ad-bed9-28c9d1ab2262)

Only thing is when there is no workflow run it adds a dash at the end and we guessed that it doesn't render when actually it is but only add the dash

![Image](https://github.com/user-attachments/assets/9577466c-4881-4644-8699-2401ef50f600)

![Image](https://github.com/user-attachments/assets/9642b7ac-49b0-4118-91c0-0f3aa0970b0c)

PR #218, opened 2025-01-18
fix: Workflow Grouping

egekocabas:
Refer to https://github.com/ls1intum/Helios/issues/163

### Description
- Fixed the issue where an extra dash was rendered unnecessarily.
- Added a message for pipeline groups with no workflow runs: "No workflow runs found."

## Previously
`Having extra dash`
Image
`If no workflowrun found, we don't show anything`
image

## After changes
`Having extra dash - -> Fixed`
image
`If no workflowrun found, we don't show anything --> Informative text added`
image


thielpa (approved):
Nice one

PR #221, opened 2025-01-19
feat: Environment list permissions client

gbanu:
* **restrict edit environments**
Screenshot 2025-01-19 at 01 08 06
Screenshot 2025-01-19 at 01 08 13
 

* **added unauthorized page**
Screenshot 2025-01-19 at 01 07 41

* **renamed 'Manage environments' button to 'Go to environments' if doesn't have edit permission**

### DIFF
--- a/client/src/app/components/environments/environment-list/environment-list-view.component.ts
+++ b/client/src/app/components/environments/environment-list/environment-list-view.component.ts
@@ -24,6 +24,7 @@ import { DeploymentStateTagComponent } from '../deployment-state-tag/deployment-
 import { LockTimeComponent } from '../lock-time/lock-time.component';
 import { KeycloakService } from '@app/core/services/keycloak/keycloak.service';
 import { CommonModule } from '@angular/common';
+import { PermissionService } from '@app/core/services/permission.service';
 
 @Component({
   selector: 'app-environment-list-view',
@@ -47,6 +48,7 @@ export class EnvironmentListViewComponent {
   private queryClient = inject(QueryClient);
   private keycloakService = inject(KeycloakService);
   private confirmationService = inject(ConfirmationService);
+  permissionService = inject(PermissionService);
 
   editable = input<boolean | undefined>();
   deployable = input<boolean | undefined>();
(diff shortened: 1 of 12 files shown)


client/src/app/pages/environment-list/environment-list.component.html
```diff
@@ -7,4 +7,4 @@
   </div>
 </app-page-heading>
 
-<app-environment-list-view [editable]="true" [hideLinkToList]="true" />
+<app-environment-list-view [editable]="permissionService.isAdmin()" [hideLinkToList]="true" />
```
egekocabas, line 10: I am a bit confused about the purpose of the `editable` field in the `environment-list-view` component. Does it indicate that the view allows `editing` (regardless of permissions), or does it signify that `editing` is only enabled if the user has the appropriate permissions (e.g., admin)?

Previously, it seemed the component always allowed `editing` (`[editable]="true"`) and handled permissions internally. With the updated code (`[editable]="permissionService.isAdmin()"`), the responsibility for permission checking appears to be shifting outside the component.

Could you clarify the intended behavior and ensure consistency in permission handling in parts where this component is used?

client/src/app/pages/unauthorized-page/unauthorized-page.component.html
egekocabas, line 1: Shouldn't it be better if we have the navigation bar in this page? like below

image

What do you think?

egekocabas (changes requested):
Thank you for the changes, i tested it and it works perfect 🚀  Left couple of points to discuss through

client/src/app/components/environments/environment-list/environment-list-view.component.ts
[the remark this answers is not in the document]
gbanu: added the permission check

client/src/app/pages/environment-list/environment-list.component.html
gbanu: moved all the permission logic inside the <app-environment-list-view/> component

client/src/app/pages/unauthorized-page/unauthorized-page.component.html
gbanu: The navigation bar needs refactoring anyways, i would tackle it in a separate pr

egekocabas:
@gbanu Can you update this PR, it doesn't have the latest `main` changes so i can not test the enabled environments
After that please use the `re-request review` button in the review section of this PR

egekocabas (approved):
Looks good, thanks for the changes 👍🏻

PR #222, opened 2025-01-19
feat: Combined Deployment and Locking History

TurkerKoc:
- Combined Deployment history with locking and unlocking history.
- Now we have Activity History on each environment 
image

- Each activity history consists of 'UNLOCKING_EVENT', 'LOCKING_EVENT' and 'DEPLOYMENT' types. 
image

- Added a check before unlocking. If there is an ongoing deployment we are not allowing user to unlock the environment. If 20 minutes passed after deployment has started, we allow to unlock again.
Screenshot 2025-01-19 at 23 13 12


client/src/app/pages/environment-deployment-history/environment-deployment-history.component.html
```diff
@@ -17,42 +34,66 @@
         </tr>
       </ng-template>
     </p-table>
-  } @else if (deploymentsQuery.data()?.length === 0) {
+  } @else if (activityHistoryQuery.data()?.length === 0) {
     <!-- Empty State -->
-    <div class="text-gray-500">No deployments found.</div>
+    <div class="text-gray-500">No deployments and locking found.</div>
```
thielpa, line 39: nitpiclk (can also be done in a separate pr): we should use similar emptystates and use the template of the table for it.
See here (from the branches-table.component.html):
```html
<ng-template pTemplate="emptymessage">
      <tr>
        <td colspan="4">
          <div class="flex flex-col gap-2 p-20 justify-center items-center">
            <i-tabler name="git-branch" class="!h-20 !w-20 text-red-500" />
            <span class="font-semibold text-xl">There are no branches found matching your filter criteria.</span>
            <span>Try clearing the filter or reloading the page.</span>
            <p-button (click)="searchTableService.clearFilter(table)">Clear Filter</p-button>
          </div>
        </td>
      </tr>
    </ng-template>
```

client/src/app/pages/environment-deployment-history/environment-deployment-history.component.html
```diff
@@ -17,42 +34,66 @@
...
-          <th>Deployed On</th>
+          <th>Type</th>
+          <th>Description</th>
+          <th>Timestamp</th>
         </tr>
       </ng-template>
-      <ng-template pTemplate="body" let-deployment>
-        <tr>
+      <ng-template pTemplate="body" let-combinedHistory>
+        <tr [class.lock-event]="combinedHistory.type === 'LOCK_EVENT'">
```
thielpa, line 51: nitpick/best-practice: You can also use the tailwind syntax here: 
`[class]="combinedHistory.type === 'LOCK_EVENT' ? 'classes-when-condition-is-true' : 'classes-when-conditions-is-false'"`

server/application-server/src/main/java/de/tum/cit/aet/helios/environment/EnvironmentLockHistoryRepository.java
```diff
@@ -52,4 +53,18 @@ Optional<EnvironmentLockHistory> findLatestLockForEnvironmentAndUser(
       """)
   Optional<EnvironmentLockHistory> findLatestLockForEnabledEnvironment(
       @Param("lockedBy") String lockedBy);
+
+  /**
+   * Returns all lock history entries for the given environment ID,
+   * sorted by lockedAt in descending order.
+   */
+  @Query("""
```
thielpa, line 61: praise: nice work!

server/application-server/src/main/java/de/tum/cit/aet/helios/deployment/ActivityHistoryDto.java
```diff
@@ -0,0 +1,52 @@
...
+
+@JsonInclude(JsonInclude.Include.NON_EMPTY)
+public record ActivityHistoryDto(
+    String type,
+    Long id,
+    RepositoryInfoDto repository,
+    Deployment.State state,
+    String sha,
+    String ref,
+    String lockedBy,
```
thielpa, line 16: question: why are we adding the lockedBy user only with its username here and not the whole user object?

thielpa (approved):
Thanks for this PR, nice work here! I just have some nitpicks, otherwise looks ready to merge from my pov!

server/application-server/src/main/java/de/tum/cit/aet/helios/deployment/ActivityHistoryDto.java
TurkerKoc: We are getting the locking user via parsing the JWT only and we are getting the preffered_username from there. It's now coming from our db.

PR #223, opened 2025-01-19
feat: Separate protected branches

gbanu:
Screenshot 2025-01-19 at 01 51 20


client/src/app/components/branches-table/branches-table.component.html
egekocabas, line 1: Would be good if we can also separate the `default` branch and list it at the top? what do you think?

image

egekocabas, line 1: Headers are shifted, can you also fix this

image

Previously, it was like below 

image

egekocabas (changes requested):
Thank you for the PR! I left some comments 👍🏻

egekocabas (approved):
Thanks for the changes, looks pretty good 👍🏻

PR #230, opened 2025-01-19
test: Add e2e test setup

thielpa:
tbd


thielpa:
Closed, so it can be reopened with staging as base

PR #215, opened 2025-01-16
test: Add server unit test setup

thielpa:
This pull request includes several changes to improve test coverage, configuration, and code quality in the `server/application-server` module. The most important changes include adding a new GitHub workflow for running tests, updating the build configuration to include test coverage reports, and adding a test case for the `BranchController` to show a example how tests are written.

### Improvements to test coverage and configuration:

* [`.github/workflows/test.yml`](diffhunk://#diff-faff1af3d8ff408964a57b2e475f69a6b7c7b71c9978cccc8f471798caac2c88R1-R44): Added a new GitHub Actions workflow to run unit and integration tests, report coverage to Codacy, and annotate test results.
* [`server/application-server/build.gradle`](diffhunk://#diff-aa1e8c60f6712196ade953ada2a0b62bc322fcd6fde1bc3d65d50f39b718e0f6R5): Updated the build configuration to include the `jacoco` plugin for test coverage reports and added a new test dependency for embedded database testing.

### Code quality improvements:

* [`server/application-server/src/main/java/de/tum/cit/aet/helios/HeliosApplication.java`](diffhunk://#diff-310942ed764fe4592766faa890f2726957124556e9445ee323677f51b862d95bL5-L10): Removed the `@EnableLoadTimeWeaving` annotation and moved it to a new configuration class.
* [`server/application-server/src/main/java/de/tum/cit/aet/helios/branch/BranchController.java`](diffhunk://#diff-f2d46cccd8187594b82e3c8bcf012bd677368f3047f7ce9d0457cc49f662f1e1L29-R29): Updated the `getBranchByRepositoryIdAndName` method to use named parameters for better readability.

### New test cases:

* [`server/application-server/src/test/java/de/tum/cit/aet/helios/branch/BranchControllerTest.java`](diffhunk://#diff-400ce1bcadae74f8b1fee9f061127832e656d42dec19ad02515ee776e452fb2bR1-R117): Added a example test cases for the `BranchController.

PR #234, opened 2025-01-20
chore: Add flyway with initial schema for migrations

StefanNemeth:
In order to support changes to our database schema in the future, a migration tool is required. Flyway and Liquibase are the most popular solutions for Spring projects. I chose Flyway as it is really simple to understand and use. It basically looks for versioned SQL files within the `server/application-server/src/main/resources/db/migration` directory and applies the ones it has not applied yet when booting the application server.

Currently we already have a database on our local environments and the production environment. So I set up baselining: On an non-empty database without Flyway, the initial migration is skipped. As an alternative, we could also drop our databases on production and locally to make sure that unused columns are deleted, etc.

I set `ddl-auto` to `validate` on the dev config, so Spring compares the current database state with what it is supposed to be, so we can catch errors here. We could also add a CI task for that, but that might be more complicated as we need a running Postgres database for that AFAIK.

To keep in mind: If we merge any other pull request that changes the database before this one, we will have to update the initial schema here or add a new one. The schema represents the state of the commit 525439b on main.


server/application-server/src/main/resources/db/migration/V1_0_0__initial_schema.sql
TurkerKoc, line 1: As you mentioned in the description. Let's be careful about changes in the main. I'm not sure whether there are some changes in the schemas. Can we reiterate on it just before merging it?

server/application-server/src/main/resources/application-prod.yml
```diff
@@ -4,11 +4,18 @@ spring:
...
+        user: ${DATASOURCE_USERNAME}
+        password: ${DATASOURCE_PASSWORD}
+        baselineOnMigrate: true
+        baseline-version: 1.0.0
     jpa:
         database: POSTGRESQL
         show-sql: false
         hibernate:
-            ddl-auto: update
+            ddl-auto: none
```
TurkerKoc, line 18: With having ddl-auto: none, I guess we won't have any errors on production if our schema doesn't match. We won't notice until some runtime errors. Soon we will create a staging server, maybe we can set it up differently in there to make sure schemas are matching!

TurkerKoc (approved):
LGTM! Thanks for putting together Flyway integration! Overall, this looks great! Thanks for adding readme as well. 

One point: We have our keycloak database in the same postgres. However, as long as our DATABASE_URL is correctly pointing helios, we won't have problems.

Please create the migration file just before merging and also resolve some conflicts! 

Can you please also create the task for adding workflow for flyway later on.

server/application-server/src/main/resources/db/migration/V1_0_0__initial_schema.sql
StefanNemeth: Sure thing!

server/application-server/src/main/resources/application-prod.yml
StefanNemeth: We could also set `ddl-auto` to `validate` for prod to make sure there won't be an issue. It will only run on startup so there won't be any performance implication AFAIK.

server/application-server/README.md
```diff
@@ -57,3 +57,9 @@ $ ./gradlew test
 ```bash
 $ ./gradlew generateOpenApiDocs
 ```
+
+### Database migrations
+
+We are using [Flyway](https://flywaydb.org/) for database migrations. When changing the database schema, create a new migration file in `src/main/resources/db/migration` with the following naming convention: `V<version>__<description>.sql`, e.g. `V1_1__create_users.sql`.
+
+The migrations are automatically applied when starting the application.
```
egekocabas, line 65: Could you clarify the versioning convention? Specifically, how should we determine when to move to a new major version (e.g., from V1 to V2) versus using subversions like V1_1 or V1_2? A quick example would be helpful for consistency in the team.

server/application-server/README.md
egekocabas, line 1: I noticed that managing version numbers in migration scripts could potentially lead to conflicts, especially when multiple PRs add migrations with the same version.

To prevent this, we could consider adding a GitHub Actions workflow to validate migration version numbers. For example

- Ensure migration filenames follow the correct naming convention (V<version>__<description>.sql)
- Check for duplicate version numbers in the repository to avoid conflicts during Flyway execution.

One specific concern is that someone might forget to sync their branch with the main branch before merging. Since GitHub handles non-conflicting changes in separate files, a version conflict might not be detected during the merge itself, leading to runtime issues later. A validation workflow could catch such cases early and ensure consistent versioning across all migrations.

Would this approach make sense? Let me know what you think about it.

egekocabas (changes requested):

server/application-server/README.md
thielpa: > I noticed that managing version numbers in migration scripts could potentially lead to conflicts, especially when multiple PRs add migrations with the same version.
> 
> To prevent this, we could consider adding a GitHub Actions workflow to validate migration version numbers. For example
> 
> * Ensure migration filenames follow the correct naming convention (V__.sql)
> * Check for duplicate version numbers in the repository to avoid conflicts during Flyway execution.
> 
> One specific concern is that someone might forget to sync their branch with the main branch before merging. Since GitHub handles non-conflicting changes in separate files, a version conflict might not be detected during the merge itself, leading to runtime issues later. A validation workflow could catch such cases early and ensure consistent versioning across all migrations.
> 
> Would this approach make sense? Let me know what you think about it.

I would suggest to just run the migration in the ci, then we cover not only the naming but also the functionality inside the migration. What do you think?

egekocabas: It sounds good, but I think we should extend this to also compare with the `main` branch to ensure no version conflicts.

For example:
- Assume the latest migration in `main` is `V1_2__update_deployment.sql`
- `PR-A` adds new migration as `V1_3__add_new_relation.sql` and gets merged into `main`
- Meanwhile, `PR-B` (which wasn’t synced with `main`) also adds `V1_3__add_roles_table.sql`
- With your proposed CI logic, it passes the CI. On the GitHub's side, this is also fine because it won’t detect a conflict since the file names are different, even though both migrations use the same version `1.3`.

I did some research, and it looks like Flyway throws an exception in this case, saying `Found more than one migration with version 1.3`

We can remind everybody to always sync their branch with the `main`, but this won’t fully prevent issues like this unless we validate the migrations in the CI.

thielpa: Yes exactly. All previous migrations should be fine, because they are validated by previous CI runs before, so we would need to run only the migration against main. Then also the error you mention is covered.

server/application-server/README.md
StefanNemeth: Hey, I clarified it. I suggested just using simple increments and documented how to deal branches here.

server/application-server/README.md
StefanNemeth: I added a Github action that applies the migrations to a Postgres database and validates them. Additionally it checks if existing migrations are removed or updated which we don't want to happen.

We could additionally add a schema validation test but that requires the merge of #215. It would check if the entities have properties that are missing in the migrations. I'd recommend we open another PR for that.

docs/development/local/migrations.rst
egekocabas, line 1: Thank you for the documentation

- This file is currently not included in the actual documentation page
please run `docker run --rm -it -v ${PWD}/docs:/docs -p 8000:8000 $(docker build -q -t sphinx -f docs/Dockerfile ./docs)` in the project root path and navigate to the `localhost:8000`
Please inspect the file [index.rst](https://github.com/ls1intum/Helios/blob/main/docs/index.rst) and folders [development](https://github.com/ls1intum/Helios/tree/main/docs/development) and [system](https://github.com/ls1intum/Helios/tree/main/docs/system). Just try to find a suitable place for this and if not exist then please create a new section.

- `Run migrations locally` -> sounds little ambiguous to me. I mean as we talked the migrations are handled automatically right. It could be better to have another sentence to also mentioning that maybe 👍🏻 
- Please include a note about the GitHub Actions workflow that validates migrations. A short explanation of how it works would be beneficial.
- `**Important**: Never modify committed migrations!` -> is great, but it would also be useful to explain what to do if a migration needs to be modified after merging. Should the author create a new migration file instead? What would be the best practice in such cases? Adding this clarification will help avoid confusion I believe.

PR #235, opened 2025-01-20
chore: Add sentry to both client and server

StefanNemeth:
Adding Sentry to our client and server to gain real-time visibility into errors and performance issues, ensuring quicker detection and resolution of problems. This will improve overall system reliability, user experience, and streamline debugging processes for both client and server components.

I also added a Github Action file that uploads the current source map of the main branch to Sentry so debugging is easier.


.github/workflows/sentry.yml
```diff
@@ -0,0 +1,41 @@
...
+      - name: Create Sentry Release and Upload Source Maps
+        uses: getsentry/action-release@f6dfa3d84a1c740b94aa45255c5e032b744a095d
+        with:
+          environment: 'production'
+          inject: true
+          sourcemaps: './client/dist/client'
+        env:
+          SENTRY_AUTH_TOKEN: ${{ secrets.SENTRY_AUTH_TOKEN }}
+          SENTRY_ORG: ${{ vars.SENTRY_ORG }}
+          SENTRY_PROJECT: ${{ vars.SENTRY_ORG }}
```
egekocabas, line 41: Is it an accident or are we using the same variable `SENTRY_ORG` for both **org** and **project**? because i also see a variable called `SENTRY_PROJECT` in the repository environments.

.github/workflows/sentry.yml
egekocabas, line 1: I have never used sentry so i actually don't know what is going on but it looks like we are sending the generated client code to sentry but do we also need to send the generated jar?

client/package.json
```diff
@@ -6,7 +6,8 @@
     "build": "ng build",
     "test": "ng test",
     "lint": "eslint --max-warnings 0 --fix .",
-    "generate:openapi": "openapi-ts"
+    "generate:openapi": "openapi-ts",
+    "sentry:sourcemaps": "sentry-cli sourcemaps inject --org helios-jf --project helios-client ./dist/client && sentry-cli sourcemaps upload --org helios-jf --project helios-client ./dist/client"
```
egekocabas, line 10: instead of hardcoded **org** and **project** can we use the environment variables?

client/src/main.ts
```diff
@@ -1,5 +1,21 @@
 import { bootstrapApplication } from '@angular/platform-browser';
 import { appConfig } from './app/app.config';
 import { AppComponent } from './app/app.component';
+import * as Sentry from '@sentry/angular';
+import { environment } from 'environments/environment';
+
+if (environment.production) {
+  Sentry.init({
+    dsn: 'https://0ef8bdd2438d4ab0333984c1c3601829@o4508675143499777.ingest.de.sentry.io/4508675194290256',
```
egekocabas, line 9: Exposing this DSN directly in a public repository doesn't seem safe. Could this allow unauthorized access or misuse? Maybe storing it in an environment variable would be a more secure approach?

client/src/main.ts
```diff
@@ -1,5 +1,21 @@
...
+    integrations: [Sentry.browserTracingIntegration(), Sentry.replayIntegration()],
+    // Tracing
+    tracesSampleRate: 1.0, //  Capture 100% of the transactions
+    // Set 'tracePropagationTargets' to control for which URLs distributed tracing should be enabled
+    tracePropagationTargets: ['localhost'],
+    // Session Replay
+    replaysSessionSampleRate: 0.1, // This sets the sample rate at 10%. You may want to change it to 100% while in development and then sample at a lower rate in production.
+    replaysOnErrorSampleRate: 1.0, // If you're not already sampling the entire session, change the sample rate to 100% when sampling sessions where errors occur.
```
egekocabas, line 17: Can you rewrite the comments in here so that we know what each property does?

server/application-server/build.gradle
```diff
@@ -134,6 +135,17 @@ openApi {
...
+  // Generates a JVM (Java, Kotlin, etc.) source bundle and uploads your source code to Sentry.
+  // This enables source context, allowing you to see your source
+  // code as part of your stack traces in Sentry.
+  includeSourceContext = true
+
+  org = "helios-jf"
+  projectName = "helios-server"
+  authToken = System.getenv("SENTRY_AUTH_TOKEN")
+}
+
```
egekocabas, line 148: What does this configuration do exactly?
Does it require a specific command to run, like `gradlew sentry`?
Does it execute during the build step or at runtime?
If it's part of the build step, do we need to set the SENTRY_AUTH_TOKEN environment variable in the production build workflow as well?

server/application-server/src/main/resources/application-prod.yml
```diff
@@ -44,3 +44,7 @@ monitoring:
     # Cooldown in minutes before running the monitoring again
     runOnStartupCooldownInMinutes: ${RUN_ON_STARTUP_COOLDOWN:15}
     repository-sync-cron: "0 0 * * * *"
+
+sentry:
+  dsn: https://a07d7c305924ed8246351dc5f11864a0@o4508675143499777.ingest.de.sentry.io/4508675150118992
```
egekocabas, line 49: Same concern with the client

> Exposing this DSN directly in a public repository doesn't seem safe. Could this allow unauthorized access or misuse? Maybe storing it in an environment variable would be a more secure approach?

server/application-server/src/main/resources/application-prod.yml
```diff
@@ -44,3 +44,7 @@ monitoring:
...
+  traces-sample-rate: 1.0
```
egekocabas, line 50: Can you also add a comment about what is this property for?

egekocabas (changes requested):
Thanks for integrating Sentry 👍🏻 

I added some comments and concerns about the changes. 

Also, for Sentry, could you invite the team so we can all access it and track issues?

client/src/main.ts
StefanNemeth: What do you mean? It's not a secret and will be public by default. The client/browser will have to know where to send the request to.

server/application-server/build.gradle
StefanNemeth: We will need to add SENTRY_AUTH_TOKEN to the environment variables of the production server. It's not supposed to run locally.

.github/workflows/sentry.yml
StefanNemeth: When building the client code, the code is optimized and minified so the users receive a small package. Sending the source map (the mapping between the optimized code and the source code) to Sentry gives us better debugging: When there is a JavaScript error, we immediately see the code that caused it

.github/workflows/sentry.yml
StefanNemeth: Good catch!

client/src/main.ts
egekocabas: I wasn’t sure how Sentry works, so I was just asking 😅  If the DSN needs to be in the client code, that makes sense since everything on the client side is public anyway. Thanks for clarifying.

server/application-server/build.gradle
egekocabas: Can you add the token in prod environment secrets and also update this [deploy_docker.yml](https://github.com/ls1intum/Helios/blob/main/.github/workflows/deploy_docker.yml#L68-L95) section?

server/application-server/src/main/resources/application-prod.yml
StefanNemeth: In this case we could potentially move it over to an environment variable, but shouldn't be a safety risk either.

server/application-server/build.gradle
egekocabas: Would be also good to mention this token in the [documentation](https://github.com/ls1intum/Helios/tree/main/docs)

client/src/main.ts
bensofficial, line 9: We got our own Sentry instance on https://sentry.ase.in.tum.de. Is there any reason why you don't use that instead of using the cloud sentry?

StefanNemeth: Will do! Thanks. I wasn't aware of our instance.

StefanNemeth:
- I simplified the setup without the source maps as they are not really required for now. Replaced the Sentry plugin with the dependency for the server.
- Added comments where it made sense
- Used an environment variable for the server DSN
- Setup two Sentry projects on our own instance and configured the client to use it

server/application-server/build.gradle
StefanNemeth: Removed this to simplify things. Now only using the dependency instead of the plugin.

client/src/main.ts
StefanNemeth: Still moved it over to the environment config file so it's cleaner :)

client/package.json
StefanNemeth: Removed it, see my comment

server/application-server/src/main/resources/application-prod.yml
StefanNemeth: Updated it to use an environment variable

server/application-server/src/main/resources/application-prod.yml
StefanNemeth: Basically enabled tracing, but removed it so only errors are reported.

egekocabas (approved):
Thanks for the PR 👍🏻  It looks good to me, couple of thoughts

- I didn't know how to test it so I just run it locally.
- Also couldn't find the project in sentry, i think it should be hidden or secret so thats fine

PR #272, opened 2025-01-24
fix: Lock issue on environment update

egekocabas:
### Motivation

Fixed a bug where updating the environment with active set to true was unintentionally clearing the lock status. The intended behavior is to clear the previous lock information in the environment entity only when the environment is being re-enabled (i.e., transitioning from disabled to enabled). This ensures that no outdated lock data is retained.


TurkerKoc (approved):
Easy one and great work! I've tested it and it works perfectly!

PR #275, opened 2025-01-25
feat: Added github id into keycloak JWT

TurkerKoc:
- Before this PR, we were using preffered_username or keycloak id for authentication purposes
- We were not able to match the logged in user with github User object.
- Added github_id into JWT, hence we can use the Github User object in necessary places.
- Already updated the realm in production and we are getting github_id in JWT payload. 
- This update is for managing local development

### !!! **Important** !!!
-  You need to delete your volume to start keycloak from scratch with updated helios-example-realm.json

### DIFF
--- a/helios-example-realm.json
+++ b/helios-example-realm.json
@@ -712,6 +712,25 @@
       "authenticationFlowBindingOverrides": {},
       "fullScopeAllowed": true,
       "nodeReRegistrationTimeout": -1,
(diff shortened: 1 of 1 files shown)


egekocabas (approved):
Tested locally and able to see my GitHub user id in the JWT, also big thanks for updating the prod config 🙏🏻

PR #278, opened 2025-01-26
fix: Expand nodes and highlight text on branch search

egekocabas:
### Motivation
The change enhances user experience by automatically expanding nodes in the tree view during branch search and highlighting the matching text.

### Description
- Implemented automatic expansion of tree nodes when a branch search matches any part of the tree.
- Added `HighlightPipe` to handle text highlighting for matched search terms efficiently.
- Included tests for the `HighlightPipe`.
- Added text highlighting for matched search terms to improve visual clarity.

### Screenshots

https://github.com/user-attachments/assets/84f7d7d2-2d6f-4c02-b545-a43579420d5c

**Question:** Should we consider highlighting only the leaf nodes instead of intermediate nodes? For example, if the search key is `feature`, we would not highlight `feature` itself but only `feature/add-new-tab` or similar leaf nodes in the tree.

### DIFF
--- a/client/src/app/components/branches-table/branches-table.component.html
+++ b/client/src/app/components/branches-table/branches-table.component.html
@@ -45,7 +45,17 @@
             <div class="flex items-center gap-2">
               <p-treeTableToggler [rowNode]="rowNode" />
 
-              <p-tag [value]="branch.name" />
+              <p-tag>
+                @for (part of (branch.name | highlight: searchTableService.searchValue()); track part.text) {
+                  <span
+                    [class.bg-yellow-200]="part.highlight"
+                    [class.font-semibold]="part.highlight"
+                    [class.rounded]="part.highlight"
+                  >
+                    {{ part.text }}
+                  </span>
+                }
+              </p-tag>
               @if (branch.isProtected) {
                 <i-tabler name="shield-half" pTooltip="Protected Branch" />
               }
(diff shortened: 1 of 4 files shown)


client/src/app/components/branches-table/branches-table.component.html
```diff
@@ -45,7 +45,15 @@
             <div class="flex items-center gap-2">
               <p-treeTableToggler [rowNode]="rowNode" />
 
-              <p-tag [value]="branch.name" />
+              <p-tag>
+                <div class="flex items-center">
+                  @for (part of branch.name | highlight: searchTableService.searchValue(); track part.text) {
+                    <span [class.bg-yellow-200]="part.highlight" [class.font-semibold]="part.highlight" [class.rounded]="part.highlight">
```
thielpa, line 51: praise: Nice one! Did not know this syntactic sugar!

client/src/app/pipes/highlight.pipe.spec.ts
```diff
@@ -0,0 +1,69 @@
+import { HighlightPipe } from './highlight.pipe';
+
+describe('HighlightPipe', () => {
```
thielpa, line 3: praise: great tests!

thielpa (approved):
Looks good! Thanks for the effort!

About the question: I guess it is fine for now as you implemented it!

PR #279, opened 2025-01-26
fix: Handle cancel case and refactor workflow label confirmation dialog

egekocabas:
### Description
- Refactored the dialog message for better user experience, adding a note section to explain the purpose of the labels.
- Fixed a bug where canceling the label change would visually update the label until a page refresh.
  - **Solution**: Stored the previous label and reverted it if the user cancels the confirmation dialog.

### Screenshots

https://github.com/user-attachments/assets/7a6abc50-e2ac-45a3-ab2d-38285e0388b6


client/src/app/pages/project-settings/project-settings.component.ts
```diff
@@ -51,7 +51,7 @@ import { SelectModule } from 'primeng/select';
 export class ProjectSettingsComponent {
   private messageService = inject(MessageService);
   private confirmationService = inject(ConfirmationService);
-  queryClient = injectQueryClient();
+  queryClient = inject(QueryClient);
```
thielpa, line 54: question: is this variable used?

client/src/app/pages/project-settings/project-settings.component.ts
```diff
@@ -62,6 +62,8 @@ export class ProjectSettingsComponent {
   // For creating a new group
   showAddGroupDialog = false;
   newGroupName = '';
+  // Store the previous label temporarily for the confirmation dialog
+  previousLabel: 'BUILD' | 'DEPLOYMENT' | 'NONE' = 'NONE';
```
thielpa, line 66: nitpick: this variable can be private. Also can't we use the type from the server enum (from openapi) here?

Also I guess this variable can also be avoided completely as well when you only use [ngModel] as one way binding and set the workflow label only in the accept function

thielpa (commented):
Got a question for this PR, but it is only a nitpick. Curious what you think about it.

client/src/app/pages/project-settings/project-settings.component.ts
egekocabas: Yes we have `invalidateQueries` after successful requests 👍🏻 
It was like `injectQueryClient();` this but since it is deprecated i only updated it as `inject(QueryClient);`

client/src/app/pages/project-settings/project-settings.component.ts
egekocabas: - I updated as private

- I think the enum types are not converted as **types** after code generation but rather directly added like below:
image
image
I checked the `hey-api` documentation and examples but couldn't find a way to generate this enum types 😞 

- I also tried one way 
```javascipt
[ngModel]="workflow.label"
(onChange)="onChangeLabel(workflow, $event.value)"
```
 with  but i think the PrimeNG <p-select> doesn't support one way binding and update the label whenever it changes

TurkerKoc (approved):
I thinks this one is ready to merge. I saw the discussion with @thielpa. @egekocabas already tried some resolutions but it works fine with the current version. It's good to go

PR #282, opened 2025-01-28
feat: Add tags

thielpa:
This PR adds the tag functionality, that can be added currently to the head commit of each branch. Also it is possible to mark releases working/broken.

[Screencast from 31.01.2025 18:20:53.webm](https://github.com/user-attachments/assets/35dd7cca-05ee-42db-bf1f-44ba527f42f9)


client/src/app/pages/tag-details/tag-details.component.html
```diff
@@ -0,0 +1,73 @@
...
+            </div>
+          }
+        </div>
+      } @else {
+        <div class="text-gray-500 text-sm uppercase font-semibold tracking-wider mb-1">No reviews yet</div>
+      }
+    </div>
+  </div>
+
+  <h3 class="text-xl mb-2">Deployments</h3>
```
StefanNemeth, line 66: I would change that to something like "Environments" or something else that signals that the list contains environments that have been deployed to

client/src/app/pages/branch-details/branch-details.component.html
```diff
@@ -12,6 +20,11 @@ <h2 class="text-2xl">{{ commit()?.message }}</h2>
       <p-skeleton width="10rem" height="2rem" />
     </div>
   }
+  <div>
+    @if (isLoggedIn()) {
```
StefanNemeth, line 24: Add isAtLeastMaintainer() here and in the backend

client/src/app/components/tag-deployment-table/tag-deployment-table.component.ts
```diff
@@ -0,0 +1,45 @@
...
+      return 'REPLACED';
+    }
+
+    if (deployments.every(deployment => deployment.state === 'SUCCESS')) {
+      return 'SUCCESS';
+    }
+
+    if (deployments.some(deployment => deployment.state === 'FAILURE')) {
+      return 'FAILURE';
+    }
```
StefanNemeth, line 42: Could you add some comments or change the var name so it's clear what's actually tested here

client/src/app/components/dialogs/tag-create/tag-create.component.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  isVisible = model.required<boolean>();
+  branchName = input.required<string>();
+  headCommit = input.required<CommitInfoDto>();
+  repositoryId = input.required<number>();
+
+  isCommitListVisible = signal(false);
+  tagName = signal('');
+
+  newCommitListQuery = injectQuery(() => getCommitsSinceLastTagOptions({ path: { name: this.branchName() } }));
```
StefanNemeth, line 34: Might want to add "enabled" here to not trigger the commit list when we don't need it

StefanNemeth (changes requested):
Good job! I added some comments.

TurkerKoc:
I found a minor bug. We experienced this issue on another controller as well. If you send the branch name as a path parameter it crashes if it contains '/' in it. It's better if we can send it in the body 👍 

image

Issue #285, opened 2025-01-30
Branch name format issues after branch search


egekocabas:
When the branch name is too long and since it is breakable, after adding the highlight then the format breaks
The issue happens when the screen is not big enough (external monitor vs laptop screen) so that the branch name doesn't fit directly to a single line

<img width="1030" alt="Image" src="https://github.com/user-attachments/assets/a8810964-101f-4aea-a4dc-b2b4f5d77f63" />

PR #286, opened 2025-01-30
fix: Branch search format breaking

egekocabas:
### Motivation
Refer to https://github.com/ls1intum/Helios/issues/285

### Screenshots
I made a change to style to prevent centering only the highlighted part when the branch name doesn't fit to the screen, after changes, it looks like below:

It doesn't feel like the current changes is the solution but it got better then the existing one. This huge branch names also causes the icons to change sizes, that is also another issue

https://github.com/user-attachments/assets/d652e5f8-a32b-4e1e-8e1a-a7e6b88e8edd


egekocabas:
I have used the mark tag but the problem i was trying to mention is not related with it actually, so i uploaded the final version can somebody check it out? @ls1intum/helios

TurkerKoc (approved):
Thanks for addressing the problems! Maybe not the best but better than the previous version. Let's fix this issue in another later on!

Issue #168, opened 2025-01-06
Tooltip for Environment Lock Counter & Redirect to Environment List


egekocabas:
I think it would be good if i hover over the timer and it says for which environment and maybe other informations as well. Also it would be good if i click on it, it redirects to the environment list page. 

![Image](https://github.com/user-attachments/assets/5ede8170-6ffd-490b-aeba-21b20c33bb95)

**Note:** Keep in mind that we can lock multiple environments right now but i think we only show the first one in the navigation bar (longest).

PR #291, opened 2025-01-31
feat: Show more details of a locked environment in tooltip

StefanNemeth:
## Motivation

After locking an environment, the user currently sees their longest locked time on the sidebar. However, it's not completely clear what that time means. Additionally, it's not possible to navigate to the environments by clicking on the time even though users might want/expect that. That's why #168 was created.

## Changes
- Move lock info of the user from the main layout component into its own
- Turn the info container into a link to the environment list, so users can unlock directly
- Add a tooltip that shows the environment that is locked and the latest deployment branch of that environment
- Add getLatestDeployment() function in environment for convenience and to avoid logic duplication
- Slight design change to show lock icon and time vertically because the sidebar got a lot wider when the timer was visible

## Screenshots

Screenshot 2025-01-31 at 21 50 05

### DIFF
--- a/client/src/app/components/user-lock-info/user-lock-info.component.html
+++ b/client/src/app/components/user-lock-info/user-lock-info.component.html
@@ -1,0 +1,19 @@
+@if (latestLock(); as lock) {
+  <a routerLink="environment/list" class="text-sm text-gray-500 flex flex-col items-center gap-1 mb-5" [pTooltip]="tooltipContent">
+    <i-tabler name="lock"></i-tabler>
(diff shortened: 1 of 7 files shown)


gbanu (approved):
Works perfect 👍 we are not planning to show multiple timers for multiple environments, right? Just needs a rebase/merge

StefanNemeth:
> Works perfect 👍 we are not planning to show multiple timers for multiple environments, right? Just needs a rebase/merge

Nope, but we could definitely add it to the backlog. Right now it shows the environment with the longest lock time.

PR #294, opened 2025-02-02
fix: LOB stream error when processing deployment

egekocabas:
### Motivation
Missing `@Transactional` was causing and exception while processing deployment

### Screenshots

- The error
image

- After the `@Transactional`
image
- The UI is also reflects
image
image


TurkerKoc (approved):
Thanks for catching this issue! LGTM

PR #297, opened 2025-02-03
feat: Implement pipeline for processing test workflow results

StefanNemeth:
I started looking into the implementation of our test case overview that we want to show within Helios in the near future for branches and PRs. Right now, we're only showing the checks/status of the workflow runs.

## Objective
This PR will cover a first implementation of a test processing pipeline that will create `TestResult` entities based on a `WorkflowRun` containing Junit test results (xml files) within its artifacts. Additionally, it should expose challenges or edge-cases that we might have to deal with in the future.

## Current pipeline
1. Artifact fetching
  - Retrieves workflow run artifacts via GitHub API (returns `PagedIterator`)
  - Looks for specifically named artifact ("Test Results")
2. ZIP processing
  - Downloads and unpacks zip archive (using stream)
  - Searches for JUnit XML files within the archive
3. Result parsing
  - Simple JUnit XML parsing using JAXB
  - Basic aggregation of totals (passed/failed/time)
5. Persistence
  - Stores summarized results linked to workflow runs

## Todos

- [ ] **Trigger Test Workflow Processing**
Implement automatic processing of a test workflow when relevant events occur (e.g., upon receiving a webhook event indicating completion or failure).

- [ ] **Consider using labels or markers to limit processing to relevant workflows only.**
Handle Invalid Files and Enforce Zip Format
Ensure proper handling of invalid files by verifying that uploaded artifacts are always in the zip format.

- [ ] **Manage Large Files**
Develop strategies for handling large files, including setting a maximum file size or applying timeouts to prevent prolonged processing.

- [ ] **Error Handling & Validation**
"Test Results" artifact could be uploaded as a non-zip, XML/JUnit reports could be malformed, etc.

- [ ] **Prevent Duplicate Artifact Processing**
Ensure that the artifacts of a workflow run are processed only once to avoid redundant operations.

- [ ] **Delay Workflow Processing**
Introduce a short delay (e.g., a few seconds) before processing completed workflows, giving the system time to make artifacts fully available.. (maybe not necessary)

- [ ] **Queue-Based Workflow Test Processing**
Implement a queue-based system for workflow test result processing to prevent blocking the webhook handler. This is crucial for handling large artifacts, which may take time to process.

- [ ] **Implement API routes**
Implement routes for fetching the latest test results given a branch or a pull request.

## Testing

I created a [repository](https://github.com/ege-test-org/test-junit-repo) that contains a simple test workflow matching the configuration that is required here (artifact containing junit file). Right now the processor is not connected to the webhook event handler yet, so testing will be documented in a later state.

## Future Work
- Extend supported parsers so we can also include client tests
- Document usage for users (e.g. they will have to upload their results with a specific artifact name and set a workflow label)
- Show data within our UI
- Parse test results in more detail (each test case individually)
- Create a PR for Artemis so the test results are uploaded in the artifact and make sure we can deal with their existing workflow

### DIFF
--- a/server/application-server/src/main/java/de/tum/cit/aet/helios/github/GitHubService.java
+++ b/server/application-server/src/main/java/de/tum/cit/aet/helios/github/GitHubService.java
@@ -22,9 +22,11 @@ import okhttp3.Request;
 import okhttp3.Request.Builder;
 import okhttp3.RequestBody;
 import okhttp3.Response;
+import org.kohsuke.github.GHArtifact;
 import org.kohsuke.github.GHOrganization;
 import org.kohsuke.github.GHRepository;
 import org.kohsuke.github.GHWorkflow;
+import org.kohsuke.github.PagedIterable;
 import org.springframework.stereotype.Service;
 
 @Service
@@ -95,6 +97,17 @@ public class GitHubService {
    */
   public GHRepository getRepository(String repoNameWithOwners) throws IOException {
     return github.getRepository(repoNameWithOwners);
+  }
+
+  /**
+   * Retrieves the artifacts for a given repository.
+   * 
+   * @param repoNameWithOwners
+   * @return the list of repositories as a PagedIterable object (not thread-safe)
+   * @throws IOException
+   */
+  public PagedIterable<GHArtifact> getWorkflowRunArtifacts(long repoId, long workflowRunId) throws IOException {
+    return github.getRepositoryById(repoId).getWorkflowRun(workflowRunId).listArtifacts();
   }
 
   /**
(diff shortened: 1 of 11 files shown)


egekocabas:
@StefanNemeth 

Thank you so much for proactively taking on this major initial task of test analysis! I have a few thoughts to share:

Artifact Fetching → It might be beneficial to apply a cutoff date here as well to ensure we always fetch the latest results (e.g., from the past week). Initially, processing test files upon receiving a webhook event might be fine, but I don’t have a strong preference on this.

File Sizes → I’m not sure how large the generated files are, but our VM has around 70-80GB of storage. If you anticipate significant growth, it might be good to talk with the admins about a potential storage increase in advance. Alternatively, if we parse the files and store only relevant data in the database, this might not be an issue—but I can't assume the storage impact just yet.

Additionally, I'm unsure about how many test libraries we should support and how many exist. Our main priority is covering all the test suites used in Artemis, but we should also consider other repositories under ls1intum.

Do you think there’s a way to uniquely identify the generated test files? Ideally, if someone wants to use Helios test analysis, they could simply select the test workflow, and we would dynamically check the generated files, detect which libraries were used, and process them accordingly based on our implemented analyzers.

I’m not deeply involved in testing, so some of my suggestions or questions might sound off—but just putting my thoughts out there! 😄

thielpa:
Closed, so it can be reopened with staging as base

PR #301, opened 2025-02-03
fix: Filter PR Fix & Pagination for activity history

TurkerKoc:
- Added Pagination for Activity History
- Filtering PR's was not working for users with username containing uppercase letters. Changed the comparison to lowercase comparison


egekocabas (approved):
👍🏻

PR #315, opened 2025-02-04
feat: Add information banner if logged out

thielpa:
If the user is not logged in,  the readonly banner is shown to ensure the user is aware of it
image


TurkerKoc (approved):
LGTM!

PR #341, opened 2025-02-04
feat: Implement pipeline for processing test workflow results

StefanNemeth:
I started looking into the implementation of our test case overview that we want to show within Helios in the near future for branches and PRs. Right now, we're only showing the checks/status of the workflow runs.

## Objective
This PR will cover a first implementation of a test processing pipeline that will create `TestSuite` and corresponding `TestCase` entities based on one or multiple JUnit Report XML files that are uploaded as artifacts by a workflow run. Additionally, I want to display the results in a simple way within the branch and PR details view.

## Current pipeline
1. Artifact fetching
  - Retrieves workflow run artifacts via GitHub API (returns `PagedIterator`)
  - Looks for specifically named artifact ("Test Results")
2. ZIP processing
  - Downloads and unpacks zip archive (using stream)
  - Searches for JUnit XML files within the archive
3. Result parsing
  - Simple JUnit XML parsing using JAXB
  - Basic aggregation of totals (passed/failed/time)
  - Parsing test cases with type (error, failure, skipped, success) and message/stackTrace
5. Persists all results to database

## Conditions
1. The label of at least one workflow is set to TEST
2. The test workflow uploads an artifact called "JUnit Test Results" containing junit xml report files
3. The test workflow is configured to be a member of a workflow group

## Screenshot/Video
https://github.com/user-attachments/assets/57ea6339-1bf0-4bea-b582-34e17c337f2e

## Additional changes
- Fixed the weird spinning animation of the workflow run icons
- Changed the environment scheduler to not run when the OpenAPI profile is active

## Testing

I created a [repository](https://github.com/ege-test-org/test-junit-repo) that contains two test workflows:
- `test-artemis.yml` uploads the results of a full Artemis test with 1 failure
- `test.yml` actually runs the tests of a Spring boot application, however there is just one test in there right now

In order to test it, you'll have to set `REPOSITORY_NAME` in your env variables to `ege-test-org/test-junit-repo`, let it sync, set the labels of `test-artemis.yml`, `test.yml` or both to "TEST" within the project settings, set up a webhook within `ege-test-org/test-junit-repo` and trigger a test run (just change a file and push to main).

## Future Work
- Extend supported parsers so we can also include client tests
- Document usage for users (e.g. they will have to upload their results with a specific artifact name and set a workflow label)
- Implement some kind of timeout for the processing of the tests + max values for the size of the artifact
- Delete old test suites + cases or just old runs in general? We might need the data for detecting flaky tests.


gbanu (approved):
Great work! Tested it for the test repo, and it works perfectly! I think we can merge it, because it doesn't introduce any breaking changes, and if we encounter some issues with the Artemis just fix them later 👍

PR #349, opened 2025-02-07
fix: Use durable consumer for GitHub webhook events

egekocabas:
### Motivation
Currently, we are using an ephemeral consumer, which causes all webhook events to be reprocessed from scratch whenever we redeploy. Switching to a durable consumer ensures event processing continuity and may help resolve the connection issues experienced during our launch. Additionally, these changes lay the foundation for future runtime updates to consumer subjects.

Furthermore, to handle cases where NATS unexpectedly removes our durable consumer (e.g., due to inactivity or other issues), we introduced a `NatsErrorListener` that detects consumer deletion errors and triggers consumer reinitialization.

### Description
- **Durable Consumer Support**
  - Added `NATS_DURABLE_CONSUMER_NAME` environment variable.
  - Updated deployment action, Docker configuration, documentation and environment variable in repository settings.
  - Updated the Docker Compose configuration to use `nats:2.10.25-alpine` instead of `nats:alpine` for consistency with production. This ensures that features like `createOrUpdateConsumer` work as expected.
- **NATS Consumer Enhancements**
  - Introduced an `updateSubjects` method that triggers `setupOrUpdateConsumer` with the latest subjects. This method is crucial for updating consuming subjects at runtime for another PR https://github.com/ls1intum/Helios/pull/351.
  - Made both `setupOrUpdateConsumer` and `updateSubjects` `synchronized` in `NatsConsumerService` to ensure thread safety during runtime subject updates.
- **Consumer Error Handling & Auto-Recovery**
  - Added `NatsErrorListener` to detect critical NATS errors, such as `Consumer Deleted (409)` errors.
  - If NATS removes the consumer, `NatsErrorListener` logs the event and calls `reinitializeConsumer` from `NatsConsumerService` to recreate the consumer automatically.
  - Addressed a circular dependency between `NatsConsumerService` and `NatsErrorListener` by using setter injection with `@Lazy` and `@Autowired` annotations to break the dependency loop while ensuring correct bean initialization.
- **Consumer Configuration Enhancements via Environment Variables**
  - `NATS_CONSUMER_INACTIVE_THRESHOLD_MINUTES` (default: 30) specifies the time (in minutes) after which an inactive consumer is removed.
  - `NATS_CONSUMER_ACK_WAIT_SECONDS` (default: 60) specifies the time (in seconds) that NATS waits for a message acknowledgment before resending the message.

### Durable Consumer Behavior and Runtime Updates
- **NATS Sequence and Durable Consumer**
  - With a durable consumer, the last processed event sequence is maintained. On redeployments, the consumer resumes from the last acknowledged sequence, ensuring that events are not reprocessed.
  - **Example**: If the consumer has processed up to sequence 100 and a redeploy occurs, it will resume processing from sequence 101 rather than reprocessing sequences 1–100.
- **Runtime Updates on Subjects**
  - The `updateSubjects` method allows the consumer to update the subjects it listens to at runtime. This is essential when the GitHub App is installed on new repositories or removed from existing ones.
  - **Important Note**: When a new subject is added, previous events for that subject will not be processed. In our case, this is acceptable because:
    - **Installation Scenario**: The GitHub App installation event establishes the starting point for that repository—there are no prior events to process.
    - **Removal Scenario**: Once the GitHub App is removed, webhook events for that repository cease.

#### Durable stream
1. Update your `.env` file under `src/application-server` and set `NATS_DURABLE_CONSUMER_NAME`.
2. Run the Docker Compose.
3. Start the **application-server** and verify that it processes events from the beginning.
4. Once processing is complete, stop the **application-server** and restart it.
5. Confirm that it does not re-process past events.
#### Reconnect (recreate) Consumer
- If doesnt matter but we will try with and without `NATS_DURABLE_CONSUMER_NAME` so repeat below steps for both
- Run compose & server
- Example log
```
Consumer created or updated with name '<you-consumer-name>' and configuration: ConsumerConfiguration { ... }
Successfully started consuming messages.
```
**Consumer Name**: <you-consumer-name> (randomly generated or durable name)
**Stream**: github
- Connect to a NATS CLI Environment
  - `docker run --rm -it --network=host natsio/nats-box`
- List Consumers in the `github` Stream
  - `nats consumer list github --server=nats://<NATS_AUTH_TOKEN>@localhost:4222`
  - Use the `NATS_AUTH_TOKEN`defined in your project root `.env` file
- Example log
```
Consumers for Stream github:

    TfGIIr1Z
    BoNZPSs9
    ...
```
Find <you-consumer-name> and see it exist.
- Delete the Consumer
  - `nats consumer delete github <you-consumer-name> --server=nats://<NATS_AUTH_TOKEN>@localhost:4222`
- Expected log output
```
[SEVERE] pullStatusError, Consumer Name: TfGIIr1Z, Status: Status{code=409, message='Consumer Deleted'}
Consumer 'TfGIIr1Z' was deleted, triggering reinitialization of consumer...
NATS Consumer reinitialization process started.
Consumer created or updated with name 'NewConsumerName' and configuration: ConsumerConfiguration { ... }
Successfully started consuming messages.
```
- It doesn't matter if your consumer if durable or ephemeral, if NATS removed the consumer and send the `Consumer Deleted` then the sequesnce data is removed from NATS and when you reinitialize your consumer it starts sending events from the beginning (config).


server/application-server/src/main/java/de/tum/cit/aet/helios/nats/NatsConsumerService.java
```diff
@@ -116,43 +136,73 @@ private Options buildNatsOptions() {
...
-                .deliverPolicy(DeliverPolicy.ByStartTime)
-                .startTime(ZonedDateTime.now().minusDays(timeframe));
+      if (existingConsumer != null) {
+        // Use existing configuration as base
+        ConsumerConfiguration existingConfig = existingConsumer.getConsumerConfiguration();
+        log.info("Consumer '{}' already exists. Updating subjects.", durableConsumerName);
+        consumerConfigBuilder = ConsumerConfiguration.builder(existingConfig)
+            .inactiveThreshold(Duration.ofMinutes(10))
+            .ackWait(Duration.ofMinutes(1))
+            .filterSubjects(subjects);
```
egekocabas, line 177: Will change those hardcoded values with environment variable

server/application-server/src/main/java/de/tum/cit/aet/helios/nats/NatsConsumerService.java
```diff
@@ -190,6 +240,21 @@ private void handleMessage(Message msg) {
...
+  /**
+   * Re-fetch subjects and re-create/update the consumer.
+   */
+  public synchronized void updateSubjects() {
+    if (natsConnection != null) {
+      try {
+        setupOrUpdateConsumer(natsConnection);
+      } catch (Exception e) {
+        log.error("Failed to update consumer at runtime: {}", e.getMessage(), e);
+        // Schedule a retry after 5 seconds delay
```
egekocabas, line 252: Will add sentry log here

server/application-server/src/main/java/de/tum/cit/aet/helios/nats/NatsConsumerService.java
```diff
@@ -116,43 +136,73 @@ private Options buildNatsOptions() {
         .build();
```
egekocabas, line 136: Also for reconnection will add log, for disconnect or removal of consumer (if we can detect it) add sentry log

server/application-server/src/main/java/de/tum/cit/aet/helios/nats/NatsConsumerService.java
```diff
@@ -190,6 +243,44 @@ private void handleMessage(Message msg) {
...
+        try {
+          setupOrUpdateConsumer(natsConnection);
+          return;
+        } catch (Exception e) {
+          log.error("Failed to update consumer at runtime: {}", e.getMessage(), e);
+          Sentry.captureException(e);
+        }
+      }
+    }
+  }
```
TurkerKoc, line 282: are these methods called by NatsClient by default?

server/application-server/src/main/java/de/tum/cit/aet/helios/nats/NatsErrorListener.java
```diff
@@ -0,0 +1,124 @@
...
+ * Custom error listener for NATS connection.
+ * Inspired by: <a href="https://natsbyexample.com/examples/os/intro/java">natsbyexample</a>
+ */
+@Log4j2
+@Component
+public class NatsErrorListener extends ErrorListenerConsoleImpl {
+
+  private final NatsConsumerService natsConsumerService;
+
+  @Autowired
```
TurkerKoc, line 28: I guess this Autowired annotation is unnecessary because Spring automatically autowires the only constructor of a bean class, even if it's not annotated. Since NatsErrorListener has only one constructor, we can remove the annotation.

TurkerKoc (approved):
Thanks for dealing with our NATS configuration. I've tested NATS as described in the description. Everything seems to work perfectly. I believe that this one will resolve our NATS heartbeat errors in prod, since it's increasing thresholds and also reinitializing the durable consumer. 

Thanks for the great work and detailed description of the PR! I've added 1-2 comments. Let's merge this one soon 🚀

server/application-server/src/main/java/de/tum/cit/aet/helios/nats/NatsErrorListener.java
egekocabas: I wasn't really sure about whether the `Autowired` should stay or not since between the `NatsErrorListener` and `NatsConsumerService` we have a **circular dependency** and all the documentaiton and articles bout solving the issue was pointing me out to also use `Autowired` with it.  

Also the javadocs of `@Lazy` annotation is also suggest me to use it with `Autowired` 
image

I mean i think when we don't use `Autowired` on the constructor and when spring was deciding on the bean initialization and injections on the startup i think if a class has only one constructor then it doesn't have to have `Autowired` annotation.

I also tested without the `Autowired` and it worked 🥳  

But i would just propose to keep it in here if that makes sense 😄

server/application-server/src/main/java/de/tum/cit/aet/helios/nats/NatsConsumerService.java
egekocabas: As we talked, i removed the other one 👍🏻

PR #351, opened 2025-02-07
feat: Dynamic repo registration

egekocabas:
> [!WARNING]
> **❗ ❗ ❗  This PR contains a migration script ❗ ❗ ❗** 

### Motivation
Enable dynamic repository syncing by accurately tracking which repositories have the GitHub App installed. This ensures our system reflects real-time changes when repositories are added or removed.

### Description
- **Webhook Listener Refactoring**
  - Updated `GitHubCustomMessageHandler` to deserialize events and delegate processing to extenders, following the `org.kohsuke:github-api` library implementation that we have.
  - Introduced `isGlobalEvent()` in both `GitHubCustomMessageHandler` and `GitHubMessageHandler` to differentiate **global events** (like `installation_repositories`) from **repository-specific** ones.
- **Subscription to global events**
  - Modified `NatsConsumerService.getSubjects` to correctly subscribe to global events using the pattern `github.?.?.installation_repositories`.
- **Repository Sync Foundation**
  - Updated `GitHubFacade` to fetch the list of repositories where the GitHub App is installed.
    - If the auth type is PAT, then it returns repositories defined in the environment variable.
- **GitHubDataSyncService Refactoring**
  - Refactored the service to sync individual repositories using the repository name with owner.
  - Added `repositoryNameWithOwner` and `status` columns to `DataSyncStatus` to maintain dedicated sync records.
  - **Did not added cascade deletion** for `DataSyncStatus` since the repository record isn’t available during the initial sync. Also it is not that important for `DataSyncStatus`.
  - Didn't make it `@Async` since it was causing so much errors related with `User` table. Further refactoring and making 'data syncing' async can be achieved if we add `repository_id` column to `User` table.
- **SyncService Class Updates**
  - Moved all the methods except the `process` methods to a new class called `GitHubDataSyncOrchestrator`. This change was necessary since we used to call these individual `process` methods inside it's own bean (`SyncService`) but this was causing **self-invocation** for the `@Transactional` annotation. 
- `GitHubInstallationRepositoriesMessageHandler`
  - Event listener for **GitHub App installed / removed to a repository** events.
  - Whenever a repository removed the GitHub App, then we remove its data from db.
  - Whenever a repository adds the GitHub App, then we start data sync for that repository.
  - At the end of this event handler, we update the subjects we listen from NATS.


server/application-server/src/main/java/de/tum/cit/aet/helios/github/sync/GitHubDataSyncScheduler.java
```diff
@@ -11,29 +17,62 @@
...
+          .filter(repo -> !syncRepositories.contains(repo))
+          .toList();
+
+      log.info("Repositories that needs to be deleted from db: {}",
+          repositoriesThatNeedsToBeDeleted);
+
+      // Sync the repositories that are installed
+      syncRepositories.forEach(dataSyncService::syncRepositoryData);
+      // Delete the repositories that are not installed anymore
+      repositoriesThatNeedsToBeDeleted.forEach(gitRepositoryRepository::deleteByNameWithOwner);
```
TurkerKoc, line 69: Let's call RepositoryService.deleteRepository here instead of directly calling repository method 👍

TurkerKoc: We also need to call DataSyncStatusService.deleteByRepositoryNameWithOwner here

server/application-server/src/main/java/de/tum/cit/aet/helios/gitrepo/github/GitHubRepositorySyncService.java
```diff
@@ -121,9 +48,10 @@ public GitRepository processRepository(GHRepository ghRepository) {
             .orElseGet(() -> repositoryConverter.convert(ghRepository));
 
     if (result == null) {
-      return null;
+      return;
     }
 
-    return gitRepoRepository.save(result);
+    gitRepoRepository.save(result);
+    gitRepoRepository.flush();
```
TurkerKoc, line 55: Let's remove this one

server/application-server/src/main/java/de/tum/cit/aet/helios/nats/NatsConsumerService.java
```diff
@@ -214,11 +221,21 @@ private void handleMessage(Message msg) {
...
       // Check if there's a custom handler for this event type
       GitHubCustomMessageHandler<?> customHandler = customHandlerRegistry.getHandler(lastPart);
       if (customHandler != null) {
-        customHandler.handleMessage(msg);
-        msg.ack();
+        if (customHandler.isGlobalEvent()) {
+          // Early ack if it is a global event
+          msg.ack();
+          customHandler.onMessage(msg);
+        } else {
```
TurkerKoc, line 234: can you explain why we are sending ack first in global events as a comment 👍

server/application-server/src/main/java/de/tum/cit/aet/helios/nats/NatsConsumerService.java
```diff
@@ -264,22 +288,118 @@ public synchronized void reinitializeConsumer() {
...
+   *
+   * <p>This method returns a subject string formatted as:
+   * <code>github.*.*.&lt;eventName&gt;</code>.</p>
+   *
+   * @param eventName the name of the event.
+   * @return a subject string for global events, e.g.
+   *      <code>github.*.*.push</code> when <code>eventName</code> is "push".
+   */
+  private String buildGlobalSubject(String eventName) {
+    return "github.*.*." + eventName;
```
TurkerKoc, line 384: In our testing session we were getting global events in following format 'github.?.?.<event_name>'
It's working fine as we tried it together but shouldn't we change this to 'github.?.?.<event_name>' ?

server/application-server/src/main/java/de/tum/cit/aet/helios/user/github/GitHubUserSyncService.java
```diff
@@ -51,35 +55,107 @@ public User syncUser(String login) {
    *
    * @param ghUser The GitHub user data to process.
    * @return The updated or newly created User entity, or {@code null} if an error occurred during
-   *     update.
+   *      update.
    */
-  @Transactional
+  @Transactional(propagation = Propagation.REQUIRES_NEW)
   public User processUser(GHUser ghUser) {
```
TurkerKoc, line 61: I think this is left from you previous async version trial. Let's convert processUser to the old version since we don't have any concurrency.

server/application-server/src/main/resources/application-dev.yml
```diff
@@ -62,7 +62,7 @@ http:
 monitoring:
     # List of repositories to monitor in the format owner/repository
     # Example: ls1intum/Helios or ls1intum/Helios,ls1intum/Artemis
-    repositories: ${REPOSITORY_NAME}
+    repositories: ${REPOSITORY_NAME:}
```
TurkerKoc, line 65: Good catch! 🚀

TurkerKoc (changes requested):
Thanks for this important update! Now it's extremely easy for anyone in ls1intum to install our Github App and start using Helios!  

We did a testings session with @egekocabas and tried locally installing/removing our dummy Github App into our dummy repo. It works like a charm 🚀 

Besides some minor issues we've talked, code also looks fantastic. Let's address them.

Also it would be great if we can add some documentation to guide repository owners to setup their repo in Helios. Here is a [discussion](https://github.com/ls1intum/Helios/issues/106#issuecomment-2603468034) for that.
Please also do not forget to mention how they should setup their deployment workflow!

Thanks for the efforts!

TurkerKoc (approved):
Great Job! Thanks for addressing al the points we've discussed 🚀

PR #360, opened 2025-02-11
feat: Add dynamic status check intervals with cooldown period and timeout handling

StefanNemeth:
## Problem
Static 120-second checks were inefficient for environments needing immediate feedback (e.g., post-deployment). However, when we set this down to like 10 seconds, we would fetch the latest status from environments that rarely change, even for the whole night.

## Overview
This PR introduces dynamic status check intervals for environments:
- Frequent checks (10s) for environments with recent status changes or deployments.
- Less frequent checks (60s) for stable environments after a cooldown period (5 minutes).
- 10-second timeout for all checks to prevent long-running jobs.
- Trigger frequent checks when deploying to environment
- Update UI so it updates the "last check" regularly


server/application-server/src/main/java/de/tum/cit/aet/helios/environment/status/EnvironmentStatusConfig.java
egekocabas, line 1: Just one minor thing, it would be better to also add those new `@Value` variables inside the yml files so that we don't forget them later on 👍🏻

PR #366, opened 2025-02-14
feat: Add delete of release candidates

thielpa:
This PR adds the possibility to delete release candidates. 

image

### DIFF
--- a/server/application-server/src/main/java/de/tum/cit/aet/helios/releasecandidate/ReleaseCandidateController.java
+++ b/server/application-server/src/main/java/de/tum/cit/aet/helios/releasecandidate/ReleaseCandidateController.java
@@ -4,6 +4,7 @@ import de.tum.cit.aet.helios.config.security.annotations.EnforceAtLeastMaintaine
 import java.util.List;
 import lombok.RequiredArgsConstructor;
 import org.springframework.http.ResponseEntity;
+import org.springframework.web.bind.annotation.DeleteMapping;
 import org.springframework.web.bind.annotation.GetMapping;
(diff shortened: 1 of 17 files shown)


server/application-server/src/main/java/de/tum/cit/aet/helios/releasecandidate/ReleaseCandidateController.java
```diff
@@ -48,4 +49,10 @@ public ResponseEntity<Void> evaluate(@PathVariable String name, @PathVariable bo
     releaseCandidateService.evaluateReleaseCandidate(name, isWorking);
     return ResponseEntity.ok().build();
   }
+
+  @DeleteMapping("/{name}")
+  public ResponseEntity<ReleaseCandidateInfoDto> deleteReleaseCandidateByName(
```
egekocabas, line 54: Should be covered with `@EnforceAtLeastMaintainer` 👍🏻

egekocabas (changes requested):
Only did the code review I will run and test it later but it looks pretty good and clean 👍🏻

server/application-server/src/main/java/de/tum/cit/aet/helios/releasecandidate/ReleaseCandidateController.java
thielpa: Nice catch :eyes: !

egekocabas (approved):
Tested locally LGTM 🚀 

(I have updated the PR with latest staging changes; while squash merging, do not forget to remove me from the commit as the co-author 👍🏻 )

PR #368, opened 2025-02-14
fix: Add repository filter to release candidate

thielpa:
Currently all release candidates from every repository are returned for a request within a repository. With this PR this is limited to only return the release candidates of the certain repository.


egekocabas (approved):
👍🏻

PR #369, opened 2025-02-14
chore: Add replay integration for sentry

thielpa:
:mirror_ball: Open for discussion: This sentry integration helps reproducing the errors and debugging it with replaying the users behaviour in the moment the bug appears. IMO this improves our capabilities and helps us fix bugs in production faster and easier. What do you think @egekocabas @StefanNemeth @TurkerKoc @gbanu? 

Find more infos here: https://sentry.io/product/session-replay/

### DIFF
--- a/client/src/main.ts
+++ b/client/src/main.ts
@@ -13,6 +13,8 @@ Sentry.init({
   integrations: [Sentry.browserTracingIntegration(), Sentry.replayIntegration()],
   // Our sample rate for tracing is 20% in production
   tracesSampleRate: environment.production ? 0.2 : 1.0,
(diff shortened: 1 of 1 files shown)


StefanNemeth (approved):
The only issue here might be data privacy and we might want to add the following config that Sentry recommends (or maybe even setting `replaysSessionSampleRate` to 0)

```
replaysSessionSampleRate: 0.1, // This sets the sample rate at 10%. You may want to change it to 100% while in development and then sample at a lower rate in production.
replaysOnErrorSampleRate: 1.0, // If you're not already sampling the entire session, change the sample rate to 100% when sampling sessions where errors occur.
```

But I'd be good with it.

PR #380, opened 2025-02-15
feat: Deployment Progress Bar

TurkerKoc:
Added progress bar for deployments into test servers

image


egekocabas (approved):
Tested with @TurkerKoc with the latest changes of staging, works great 👍🏻

PR #388, opened 2025-02-18
fix: Deployment progress bar step transitions

TurkerKoc:
Transition between pre-deployment and deployment was not properly handled. Fixed the issue with our new Deployment Type


egekocabas (approved):
LGTM

PR #426, opened 2025-02-20
chore: Introduce git lfs

thielpa:
To reduce fetching size, i would like to introduce git lfs (https://git-lfs.com/) to replace binaries with pointers in git context.
This should not change (anything) in the use of github for you, except you maybe have to enable it on your machine with `git lfs install`. You can add new files using `git lfs track "xxx"`.

Currently I included *.jar, *.png and *.jpg files


TurkerKoc (approved):
Thanks for adding this one to our repo 🚀 
note for mac users: install git-lfs using
`brew install git-lfs`

Then you can do
`git lfs install`

PR #433, opened 2025-02-23
feat: Add pinning for prs and branches

thielpa:
This Pull request adds the possibility to pin prs/branches at the top of the table. This user preference is also persisted in the db.

[Screencast from 23.02.2025 17:11:04.webm](https://github.com/user-attachments/assets/79dccec7-2472-4743-bbbd-c6393bdfe120)


server/application-server/src/main/resources/db/migration/V15__add_user_preference.sql
egekocabas, line 1: Delete on cascade should be added, so that whenever a branch is removed, its preference is removed from preference table as well 👍🏻  

```
Caused by: org.hibernate.exception.ConstraintViolationException: could not execute statement [ERROR: update or delete on table "branch" violates foreign key constraint "fkgyxf2ri6f3j4pourba7shdsma" on table "user_preference_favourite_branches"
  Detail: Key (repository_id, name)=(890885508, feat/ege-test3) is still referenced from table "user_preference_favourite_branches".] [delete from branch where name=? and repository_id=?]

```

server/application-server/src/main/java/de/tum/cit/aet/helios/pullrequest/PullRequestService.java
```diff
@@ -1,24 +1,41 @@
...
-
-  public PullRequestService(PullRequestRepository pullRequestRepository) {
-    this.pullRequestRepository = pullRequestRepository;
-  }
+  private final UserPreferenceRepository userPreferenceRepository;
+  private final AuthService authService;
 
   public List<PullRequestBaseInfoDto> getAllPullRequests() {
+    final Optional<UserPreference> userPreference =
+        userPreferenceRepository.findByUser(authService.getUserFromGithubId());
```
egekocabas, line 26: Also the same thing with `getAllBranches`

client/src/app/components/branches-table/branches-table.component.html
egekocabas, line 1: We already talked about this branch table's tree structure and visibility of pinned branches inside the tree.
Possible solution is having another separate group called **Pinned** and we show all the branches in the tree view (nothing will be changed there) but also inside the **Pinned** group we will show all the pinned branches there again

egekocabas (changes requested):
Made a review together and come up with couple things, but other than that looks good, thanks 🍾

egekocabas (approved):
Tested locally and it works great, thanks for implementing this feature 🚀

PR #434, opened 2025-02-23
chore: Fix codacy exclude paths

thielpa:
The exclude paths from codacy do not work correctly. Hopefully this will work. I removed the ** prefix, because that is not documented in codacy to work.


TurkerKoc (approved):
LGTM! let's try it out

thielpa:
Please do not merge yet, I currently check out another fix

PR #441, opened 2025-02-25
feat: Add gh release integration

thielpa:


server/application-server/src/main/java/de/tum/cit/aet/helios/releaseinfo/ReleaseInfoDetailsDto.java
```diff
@@ -1,21 +1,24 @@
...
+public record ReleaseInfoDetailsDto(
     @NonNull String name,
     @NonNull CommitInfoDto commit,
-    @NonNull BranchInfoDto branch,
+    BranchInfoDto branch,
     @NonNull List<ReleaseCandidateDeploymentDto> deployments,
     @NonNull List<ReleaseCandidateEvaluationDto> evaluations,
-    @NonNull UserInfoDto createdBy,
+    ReleaseDto release,
+    UserInfoDto createdBy,
```
gbanu, line 21: @thielpa are you still working on this pr? I get some type issues in the client because `createdBy` and `branch` can be undefined 😕

thielpa: Hmm strange, I'll look into it

server/application-server/src/main/java/de/tum/cit/aet/helios/deployment/DeployRequest.java
```diff
@@ -3,4 +3,4 @@
 import org.springframework.lang.NonNull;
 
 public record DeployRequest(
-    @NonNull Long environmentId, @NonNull String branchName, String commitSha) {}
+    @NonNull Long environmentId, String branchName, @NonNull String commitSha) {}
```
gbanu, line 6: have you tested what happens if branchName is null?
the testserver-deplyoment workflow uses branch name
for the releases maybe we can find a branch it was created from?

server/application-server/src/main/java/de/tum/cit/aet/helios/github/sync/GitHubDataSyncOrchestrator.java
```diff
@@ -357,15 +353,37 @@ public List<GHCommit> syncCommitsOfRepository(GHRepository repository) {
...
+   */
+  public void syncReleasesOfRepository(GHRepository repository) {
+    try {
+      var iterator = repository.listReleases().withPageSize(100).iterator();
+
+      while (iterator.hasNext()) {
+        var ghReleases = iterator.nextPage();
+        // Only process non-draft releases, because drafts are not unique in GitHub
+        ghReleases.stream()
+            .filter(Predicate.not(GHRelease::isDraft))
```
gbanu, line 369: just a question:
why do we filter it out here? because we have isDraft property of a Release entity.
and also when we create a release draft with helios, we then also process it, right?

gbanu (changes requested):
Sorry for keeping this PR in review for so long, I left a couple of comments, but I somehow still can't start the client locally 😓
maybe I can try again after rebase 🙏

server/application-server/src/main/java/de/tum/cit/aet/helios/deployment/DeployRequest.java
thielpa: Actually we unfortunately cannot do that. IMO the testserver-deployment should also care more about the commitSha instead of the branch, because this commit will be deployed and the branch might not been known or it is simply an detached commit. I also modified the deployToEnvironment logic a bit to fix this issue. Maybe you could take a look into that as well and check if it makes sense, thanks!

server/application-server/src/main/java/de/tum/cit/aet/helios/releaseinfo/release/github/GitHubReleaseMessageHandler.java
```diff
@@ -0,0 +1,44 @@
...
+    super(GHEventPayload.Release.class);
+    this.releaseSyncService = releaseSyncService;
+    this.repositorySyncService = repositorySyncService;
+  }
+
+  @Override
+  protected void handleEvent(GHEventPayload.Release eventPayload) {
+    if (eventPayload.getAction().equals("created")) {
+      return;
+    }
```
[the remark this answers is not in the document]
thielpa: No, I only want to process the change and publish events here, this is why I excluded the created event. But I will change the code as you are right this is not readable enough. :pray:

server/application-server/src/main/java/de/tum/cit/aet/helios/releaseinfo/release/github/GitHubReleaseMessageHandler.java
```diff
@@ -0,0 +1,44 @@
...
+    log.info(
+        "Received release event for repository: {}, release: {}, action: {}",
+        eventPayload.getRepository().getFullName(),
+        eventPayload.getRelease().getName(),
+        eventPayload.getAction());
+    repositorySyncService.processRepository(eventPayload.getRepository());
+    // We don't need to handle the deleted action here, as pull requests are not deleted
```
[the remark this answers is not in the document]
thielpa: See comment below :)

server/application-server/src/main/java/de/tum/cit/aet/helios/releaseinfo/ReleaseInfoListDto.java
```diff
@@ -0,0 +1,14 @@
+package de.tum.cit.aet.helios.releaseinfo;
+
+import de.tum.cit.aet.helios.releaseinfo.releasecandidate.ReleaseCandidate;
+
+public record ReleaseInfoListDto(
```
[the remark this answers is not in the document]
thielpa: I also thought long about the names, but could not figure out a good one that I was fully satisfied with.
IMO we cannot use ReleaseCandidate... because this object both contains Info from the Release (of GH) and the Helios ReleaseCandidate objects.
What do you think, do you have another suggestion?

server/application-server/src/main/java/de/tum/cit/aet/helios/releaseinfo/release/github/GitHubReleaseMessageHandler.java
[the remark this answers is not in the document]
thielpa: Actually we need to process the create event as well, I misunderstoood the docs of GH, good catch here!

gbanu (changes requested):

server/application-server/src/main/java/de/tum/cit/aet/helios/releaseinfo/release/github/GitHubReleaseSyncService.java
```diff
@@ -0,0 +1,100 @@
...
+        .findByRepositoryRepositoryIdAndName(repository.getRepositoryId(), ghRelease.getTagName())
+        .map(
+            releaseCandidate -> {
+              releaseCandidate.setRelease(result);
+              return releaseCandidateRepository.save(releaseCandidate);
+            })
+        .orElseGet(
+            () -> {
+              try {
+                final GHRef ref = ghRepository.getRef("tags/" + ghRelease.getTagName());
```
gbanu, line 64: I don't know if i didn't setup something correctly, but i get
```2025-03-28T17:42:53.435+01:00 ERROR 40453 --- [Helios] [         nats:4] d.t.c.a.h.r.r.g.GitHubReleaseSyncService : Failed to get ref for release candidate 1.1.0: Offline```
and the releases from GH are not shown in Helios

gbanu (approved):

PR #444, opened 2025-02-25
feat: Extend lock logic

TurkerKoc:
#### Prerequisites:
- ...
#### Flow:
2. Navigate to Settings
3. ...

#### General
- [ ] PR description explains the purpose and changes. [(f.e. following the 'Conventional Commits')](https://www.conventionalcommits.org/en/v1.0.0/)
- [ ] Changes have been tested locally.
- [ ] Screenshots have been attached.

#### Server
- [ ] Code is performant and follows best practices
- [ ] I documented the Java code using JavaDoc style.

#### Client
- [ ] I documented the TypeScript code using JSDoc style.
- [ ] I added multiple screenshots/screencasts of my UI changes.
- [ ] I translated all newly inserted strings into English and German.

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -25,7 +25,7 @@
     "@sentry/cli": "^2.40.0",
     "@tanstack/angular-query-experimental": "^5.62.8",
     "angular-tabler-icons": "^3.1.0",
-    "eslint-config-prettier": "^9.1.0",
+    "eslint-config-prettier": "^10.0.0",
     "keycloak-js": "^26.0.7",
     "marked": "^15.0.3",
     "primeicons": "^7.0.0",
--- a/client/yarn.lock
+++ b/client/yarn.lock
@@ -4448,10 +4448,10 @@ escape-string-regexp@^4.0.0:
   resolved "https://registry.yarnpkg.com/escape-string-regexp/-/escape-string-regexp-4.0.0.tgz#14ba83a5d373e3d311e5afca29cf5bfad965bf34"
   integrity sha512-TtpcNJ3XAzx3Gq8sWRzJaVajRs0uVxA2YAkdb1jm2YkPz4G6egUFAyA3n5vtEIZefPk5Wa4UXbKuS5fKkJWdgA==
 
-eslint-config-prettier@^9.1.0:
-  version "9.1.0"
-  resolved "https://registry.yarnpkg.com/eslint-config-prettier/-/eslint-config-prettier-9.1.0.tgz#31af3d94578645966c082fcb71a5846d3c94867f"
-  integrity sha512-NSWl5BFQWEPi1j4TjVNItzYV7dZXZ+wP6I6ZhrBGpChQhZRUaElihE9uRRkcbRnNb76UMKDF3r+WTmNcGPKsqw==
+eslint-config-prettier@^10.0.0:
+  version "10.0.1"
+  resolved "https://registry.yarnpkg.com/eslint-config-prettier/-/eslint-config-prettier-10.0.1.tgz#fbb03bfc8db0651df9ce4e8b7150d11c5fe3addf"
+  integrity sha512-lZBts941cyJyeaooiKxAtzoPHTN+GbQTJFAIdQbRhA4/8whaAraEh47Whw/ZFfrjNSnlAxqfm9i0XVAEkULjCw==
 
 eslint-import-resolver-node@^0.3.9:
   version "0.3.9"


server/application-server/src/main/java/de/tum/cit/aet/helios/environment/EnvironmentService.java
```diff
@@ -235,7 +235,7 @@ private OffsetDateTime getLockWillExpireAt(Environment environment) {
   }
 
   @Transactional
-  private OffsetDateTime getLockReservationExpiresAt(Environment environment) {
+  protected OffsetDateTime getLockReservationExpiresAt(Environment environment) {
```
thielpa, line 238: question: Why can't this be private?

server/application-server/src/main/java/de/tum/cit/aet/helios/environment/EnvironmentController.java
```diff
@@ -74,7 +74,18 @@ public ResponseEntity<?> unlockEnvironment(@PathVariable Long id) {
...
+        .orElse(ResponseEntity.status(403).build());
+  }
+
+  @EnforceAtLeastWritePermission
+  @PutMapping("/{id}/extend-lock")
+  public ResponseEntity<?> extendEnvironmentLock(@PathVariable Long id) {
+    Optional<Environment> environment = environmentService.extendEnvironmentLock(id);
+    return environment
+        .map(env -> ResponseEntity.ok(EnvironmentDto.fromEnvironment(env)))
+        .orElse(ResponseEntity.status(403).build());
```
thielpa, line 88: issue: can we throw a EntityNotFoundException with a message instead? This way we miss the client notification for the user on an error right?

server/application-server/src/main/java/de/tum/cit/aet/helios/environment/EnvironmentController.java
```diff
@@ -74,7 +74,18 @@ public ResponseEntity<?> unlockEnvironment(@PathVariable Long id) {
   @PutMapping("/{id}/lock")
   public ResponseEntity<?> lockEnvironment(@PathVariable Long id) {
     Optional<Environment> environment = environmentService.lockEnvironment(id);
-    return ResponseEntity.ok(EnvironmentDto.fromEnvironment(environment.get()));
+    return environment
+        .map(env -> ResponseEntity.ok(EnvironmentDto.fromEnvironment(env)))
+        .orElse(ResponseEntity.status(403).build());
```
thielpa, line 79: issue: can we throw a EntityNotFoundException with a message instead? This way we miss the client notification for the user on an error right?

server/application-server/src/main/java/de/tum/cit/aet/helios/environment/EnvironmentService.java
```diff
@@ -253,6 +253,95 @@ private OffsetDateTime getLockReservationExpiresAt(Environment environment) {
...
+      throw new IllegalStateException("Environment is disabled");
+    }
+
+    if (environment.getType() != Environment.Type.TEST) {
+      throw new IllegalStateException("Only TEST environments can have their locks extended");
+    }
+
+    // Verify the environment is locked and by the current user
+    if (!environment.isLocked() || !currentUser.equals(environment.getLockedBy())) {
+      return Optional.empty();
```
thielpa, line 291: We can also throw the error here instead of in the controller.

server/application-server/src/main/java/de/tum/cit/aet/helios/environment/EnvironmentService.java
egekocabas: It was because of this below IDE suggestion
image

A broader issue we need to address with `@Transactional` is separating transactional methods from their calling beans. If a bean (e.g., a class) calls its own method annotated with `@Transactional`, this leads to what is known as the **Self-Invocation Problem**. In such cases, Spring does not recognize the transaction boundary, preventing proper rollback. To avoid this, we should identify transactional methods and refactor them to ensure they are invoked through a properly managed Spring proxy, eliminating self-invocation issues.

Thats why IntelliJ gives it red and let us remember to make Transactional protected/public because it should be called from another bean 😅

server/application-server/src/main/java/de/tum/cit/aet/helios/environment/EnvironmentService.java
```diff
@@ -253,6 +259,125 @@ private OffsetDateTime getLockReservationExpiresAt(Environment environment) {
...
+    if (!environment.isLocked()) {
+      throw new EnvironmentException("Environment is not locked. Cannot extend lock.");
+    } else if (!environment.getLockedBy().equals(currentUser)) {
+      // The Environment is locked, but by someone else
+      String msg = getLockedByAnotherUserErrorMessage(environment);
+      throw new EnvironmentException(msg);
+    }
+
+    // Calculate new expiration times based on current time
+    environment.setLockWillExpireAt(calculateNewLockExpiration(environment));
```
thielpa, line 305: question: Why does this have to be handled differently than on line 204/205?

thielpa (changes requested):
Nice work with this PR :100: . I just found out one unhandled error that should be fixed before merging.

The methods in the EnvironmentService are also a good starting point for writing some unit tests, but you dont have to if you rather want to skip it ;)

server/application-server/src/main/java/de/tum/cit/aet/helios/environment/EnvironmentService.java
```diff
@@ -253,6 +259,125 @@ private OffsetDateTime getLockReservationExpiresAt(Environment environment) {
...
+    environment.setLockReservationExpiresAt(calculateNewReservationExpiration(environment));
+
+    try {
+      environmentRepository.save(environment);
+    } catch (OptimisticLockingFailureException e) {
+      // The environment was modified by another transaction
+      return Optional.empty();
```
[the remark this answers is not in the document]
TurkerKoc: OptimisticLockingFailureException if for catching concurrency errors. If we catch this it means that this logic already happened and no need to throw an error. 

Information from Spring:
The OptimisticLockingFailureException is an exception that occurs in Spring when an optimistic locking conflict is detected. Optimistic locking is a concurrency control mechanism used to ensure that data is not overwritten by concurrent transactions.

server/application-server/src/main/java/de/tum/cit/aet/helios/environment/EnvironmentService.java
```diff
@@ -253,6 +259,125 @@ private OffsetDateTime getLockReservationExpiresAt(Environment environment) {
...
+    }
+
+    return Optional.of(environment);
+  }
+
+  @NotNull
+  private static String getLockedByAnotherUserErrorMessage(Environment environment) {
+    String msg = "Environment is locked by another user.\n";
+    ZoneId localZone = ZoneId.systemDefault();
```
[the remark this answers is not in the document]
TurkerKoc: Basically everything can be made final :)

server/application-server/src/main/java/de/tum/cit/aet/helios/environment/EnvironmentService.java
TurkerKoc: They are using different baseTimes, one is using locking time the other one is current timestamp. also some check are different but you're right we have duplicated code here. Will update it to reuse the duplicated parts

server/application-server/src/main/java/de/tum/cit/aet/helios/environment/EnvironmentController.java
```diff
@@ -77,6 +77,15 @@ public ResponseEntity<?> lockEnvironment(@PathVariable Long id) {
     return ResponseEntity.ok(EnvironmentDto.fromEnvironment(environment.get()));
   }
 
+  @EnforceAtLeastWritePermission
+  @PutMapping("/{id}/extend-lock")
+  public ResponseEntity<?> extendEnvironmentLock(@PathVariable Long id) {
+    Optional<Environment> environment = environmentService.extendEnvironmentLock(id);
+    return environment
+        .map(env -> ResponseEntity.ok(EnvironmentDto.fromEnvironment(env)))
+        .orElse(ResponseEntity.notFound().build());
```
thielpa, line 86: issue: this afaik still provides no message to the user, not sure if this can happen though. I actually would remove the optional from the service and return only the object instead as we are not using it anyways

thielpa (approved):
Left a small comment that would be nice if you can address it, but otherwise it looks smooth to me :)

egekocabas (approved):
Tested with @gbanu and its working, will merge it for the release and we can refactor for [what paul suggested](https://github.com/ls1intum/Helios/pull/444#pullrequestreview-2658524040) in another PR 👍🏻

PR #451, opened 2025-02-26
fix: enable deployments on phone

thielpa:
This Pull request should enable users to deploy using their phone adding responsive design to Helios. This PR does only hotfix the most basic functionality and does not focus on UX or good looking design for phones


egekocabas (approved):
Thanks for the initial mobile update! It will be easier to use on mobile in the future 👍🏻

PR #467, opened 2025-03-03
chore: set pr title validation to follow conventional commits

thielpa:
Currently it is not allowed to have a lowercase letter after the `:`, but this does not align with the conventional commit spec: https://www.conventionalcommits.org/en/v1.0.0/

I changed the gh action to a better one that has this functionality without setting a custom regex.


gbanu (approved):
maybe we can also add `style` as a task type if we are only doing some minor ui changes, but it is not as important, and we can just label it as a `feat`

PR #475, opened 2025-03-04
feat: Add behind by, ahead by and commits info for release candidate creation

### LINKED ISSUE #309: [Bug]:  Iterate over new commits in tag creation
### Contact Details

_No response_

### What happened?

If you create a tag on an older commit than the previous tag, the new commits message in the create-tag dialog is not correct, but showing "creating with x new commits"

### Steps To Reproduce

_No response_

_No response_

### Relevant log output

```shell

```

StefanNemeth:
### Motivation
Right now, we always show "x new commits" when creating a new release candidate even when the release candidate is behind the latest one. Fixes #309 

### Description
Adds the behind/ahead info and shows a few commits that were added so the user knows what would be added by the release candidate. We show a maximum of 3 commits and then show "and .. more" which links to the GitHub comparison page.

### Testing Instructions

You can test 5 different kinds of state.

- There is no previous RC (release candidate)
- A branch is identical to this one (in this case there shouldn't even be a "create release candidate" button.
- A branch is just ahead one or more commits compared to the latest RC
- A branch is behind one or more commits compared to the latest RC
- The branches diverged: The branch is ahead one or more and behind one or more commits compared to the latest RC

Just create a release candidate for a branch, then push another branch with the state you want to test and create a release candidate for that one.

### Screenshots
image

#### General


thielpa (approved):
Nice and clean. LGTM!

Maybe when we implement the pagination stuff, we should also take a look at the commits here as in Artemis they diverge by 100+ in some branches

PR #492, opened 2025-03-04
refactor: use required args constructor annotation

thielpa:
Make use of the `@RequiredArgsConstructor` to reduce code lines and also error proneness


StefanNemeth (approved):
Nice, removing 103 lines of code feels good!

PR #493, opened 2025-03-05
chore: add release pipeline for sentry

thielpa:


.github/workflows/prod.yml
```diff
@@ -1,23 +1,8 @@
...
-      - 'helios-example-realm.json'
-      - 'LICENSE'
-      - 'nginx.conf'
-      - 'openapitools.json'
-      - 'README.md'
-      - '.github/**'
-      - '.vscode/**'
-      - 'docs/**'
-      - 'keycloakify/**'
+  release:
```
TurkerKoc, line 4: Great that we are starting to deploy by creating a release. Thanks 🚀

.github/workflows/release-sentry.yml
```diff
@@ -0,0 +1,57 @@
...
+
+jobs:
+  release-client-sentry:
+    runs-on: ubuntu-latest
+    steps: 
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - name: Run
+        uses: tj-actions/docker-run@68424e50d5326da41f9197f31797493cb6d7c917
```
TurkerKoc, line 18: Using an 3rd party action is not always reliable. However, if you checked to repo and code, fine with me!

.github/workflows/release-sentry.yml
```diff
@@ -0,0 +1,57 @@
...
+        id: docker-run
+        with:
+          image: ghcr.io/ls1intum/helios/client:${{ inputs.client_image_tag }}
+          name: client-build
+          options: -d
+          args: sleep 10
```
TurkerKoc, line 24: make sure that this 10 seconds of sleep is enough. Should be fine!

.github/workflows/release-sentry.yml
```diff
@@ -0,0 +1,57 @@
...
+      - name: Copy from container to host
+        uses: tj-actions/docker-cp@d604bf259c630f3d7c8da3b39eb8a31c9519da7a
```
TurkerKoc, line 26: again 3rd party thing

client/angular.json
```diff
@@ -51,7 +51,9 @@
                 }
               ],
               "outputHashing": "all",
-              "sourceMap": true
+              "sourceMap": {
```
TurkerKoc, line 54: 👍

TurkerKoc (approved):
I did a code review. Didin't run it (since it's hard to test it like this). I've added some comments which we need to consider. Other than that everything looks fine. We can do some changes if need to later on! Great work thank 🙏

PR #497, opened 2025-03-08
chore: add build step for keycloakify theme

thielpa:
To avoid that after a dependency update or similar the build of the keycloak theme is forgotten, this pr adds an automatic build for the keycloakify theme and adds the files to the keycloak image


StefanNemeth (approved):
LGTM! :)

egekocabas (changes requested):
Great work 👍🏻 

- Could you update the deployment file so that we can use the `KEYCLOAK_IMAGE_TAG` variable. Couple of files that its need to change:
  - [deploy_docker.yml](https://github.com/ls1intum/Helios/blob/staging/.github/workflows/deploy_docker.yml#L103-L105)
  - [prod.yml](https://github.com/ls1intum/Helios/blob/staging/.github/workflows/prod.yml#L33-L41)
  - [staging.yml](https://github.com/ls1intum/Helios/blob/staging/.github/workflows/staging.yml)

egekocabas (approved):
Perfect 🙏🏻

PR #512, opened 2025-03-10
chore: Use multi-project Gradle setup and refactor NATS message handler

StefanNemeth:
### Motivation
Right now, most of our logic is within the application-server. When we want to seperate our test processor logic into its own service, we might want to share some code and dependencies (like common code). Additionally, we need to establish a form of communication among the services.

### Description
In order to be able to share common code (e.g. for the data objects) among our microservices, I changed our Gradle structure to support multiple projects. That's why, now, the app server is started within the `server` directory using `./gradlew :application-server:bootRunDev`. I changed the corresponding docs and Github actions as well.

Additionally, I refactored our NATS communication handler so it can also be used for non-Github messages and is a bit cleaner. That way, the services can use NATS (+ some common code) as a way to communicate with each other.

### Testing Instructions
General development: Make sure that running the app server, building, generating OpenAPI docs, etc. still works on your machine. Also some Github actions, like tests, flyway validate and the deployment should be tested.

#### General


thielpa (approved):
Nice one, really great job in generalizing all of this and reducing the code duplication!

PR #537, opened 2025-03-15
chore: optimize Webhook Listener's dependencies

egekocabas:
### Motivation
Simplifies `requirements.txt` by removing unnecessary dependencies. Keeps only essential packages required for the application's functionality.


gbanu (approved):
Works perfectly! 

Thanks! We will be having so much less renovate updates 😍 🎉

Issue #317, opened 2025-02-04
Tabs are not updating on route change


thielpa:
Reproduce:

Go to PR view.
Go to Branch tab.
On the sidenav click on the CI-CD icon

-> You will end up on the pr view, but the tabs still show the branches

PR #540, opened 2025-03-15
feat: link pr in the env view

gbanu:
I didn't want to overload the environment list component even more that's why added a link to the existing latest deployment tag.

https://github.com/user-attachments/assets/eeac0b13-fa36-4bda-a916-d17f2edcd72f

fixes https://github.com/ls1intum/Helios/issues/317

### DIFF
--- a/client/src/app/components/environments/environment-list/environment-list-view.component.html
+++ b/client/src/app/components/environments/environment-list/environment-list-view.component.html
@@ -44,7 +44,14 @@
                   <span [pTooltip]="getDeploymentTime(environment) || ''">{{ environment.latestDeployment.updatedAt || '' | timeAgo }}</span>
                 }
                 <app-deployment-state-tag [state]="deployment.state" />
-                <p-tag severity="secondary" rounded="true" class="max-w-[350px] flex items-center" [pTooltip]="tooltipTemplate" tooltipPosition="top">
+                <p-tag
+                  severity="secondary"
+                  rounded="true"
+                  class="max-w-[350px] flex items-center"
+                  [pTooltip]="tooltipTemplate"
+                  tooltipPosition="top"
+                  [routerLink]="getPrOrBranchLink(environment)"
+                >
                   <i-tabler name="git-branch" class="!size-5 mr-0.5 flex-shrink-0"></i-tabler>
                   <span class="ml-1 truncate flex-1">{{ deployment.ref }}</span>
                   <ng-template #tooltipTemplate>
--- a/client/src/app/components/environments/environment-list/environment-list-view.component.html
+++ b/client/src/app/components/environments/environment-list/environment-list-view.component.html
@@ -74,7 +74,7 @@
                     [routerLink]="getPrLink(environment)"
                   >
                     <i-tabler name="git-pull-request" class="!size-5 mr-0.5 flex-shrink-0"></i-tabler>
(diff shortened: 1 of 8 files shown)


egekocabas (commented):
I believe the current behavior isn't ideal for user experience. From a user's perspective, clicking on the branch name should navigate directly to the branch details. One potential solution might be to introduce an additional tag showing only the PR number (e.g., `#10254`). However, in any case, we should also include a hover effect to clarify that the element is clickable and redirects the user elsewhere.

What do you think about that? I know there are already a lot of elements in this view, but I believe this approach would provide a better overall user experience.

gbanu:
@egekocabas added a separate tag for PR and disabled navigation for release tags, Paul has implemented the release sync from github and we can implement a link to the release once his pr is merged.


https://github.com/user-attachments/assets/8c8e176b-3f47-4dab-983b-09c16f36ea8e

TurkerKoc:
> @egekocabas added a separate tag for PR and disabled navigation for release tags, Paul has implemented the release sync from github and we can implement a link to the release once his pr is merged.
> 
>  Screen.Recording.2025-03-16.at.17.23.40.mov

That looks great thanks for adding separate link for PRs!

Maybe a minor change: I think we don't need to write PR as a prefix since this is a known pattern for developers -> #<PR_NUMBER>. Also we have the icon as well :)

client/src/app/components/environments/environment-list/environment-list-view.component.html
```diff
@@ -44,13 +44,39 @@
                   <span [pTooltip]="getDeploymentTime(environment) || ''">{{ environment.latestDeployment.updatedAt || '' | timeAgo }}</span>
                 }
                 <app-deployment-state-tag [state]="deployment.state" />
-                <p-tag severity="secondary" rounded="true" class="max-w-[350px] flex items-center" [pTooltip]="tooltipTemplate" tooltipPosition="top">
-                  <i-tabler name="git-branch" class="!size-5 mr-0.5 flex-shrink-0"></i-tabler>
+
+                <!-- Branch or Release Tag -->
+                <p-tag
+                  severity="secondary"
```
thielpa, line 50: issue: please add cursor-pointer class here

client/src/app/components/environments/environment-list/environment-list-view.component.html
```diff
@@ -44,13 +44,39 @@
...
+                    <i-tabler name="git-branch" class="!size-5 mr-0.5 flex-shrink-0"></i-tabler>
+                  }
                   <span class="ml-1 truncate flex-1">{{ deployment.ref }}</span>
-                  <ng-template #tooltipTemplate>
-                    <div class="flex items-center">{{ deployment.ref }}</div>
-                  </ng-template>
                 </p-tag>
+                @if (deployment.pullRequestNumber) {
+                  <!-- PR Tag -->
+                  <p-tag
```
thielpa, line 67: issue: please also add cursor-pointer class here

thielpa (changes requested):
Nice and clean! I just have a minor change, that we should address :) :partying_face:

client/src/app/components/environments/environment-list/environment-list-view.component.html
gbanu: it is actually enabled on the whole element but was not showing during screen recording 😅 
IMG_5425
IMG_5426

TurkerKoc (approved):
LGTM! Great job, thanks for adding this feature 🚀 🎊

PR #545, opened 2025-03-16
chore: enable feature preview for keycloak

gbanu:
In order to be able to configure the keycloak for user impersonation we have to start the container with the flag `--features="token-exchange,admin-fine-grained-authz"`

### DIFF
--- a/compose.prod.yaml
+++ b/compose.prod.yaml
@@ -151,7 +151,7 @@ services:
       - postgres
     volumes:
       - ./helios-realm.json:/opt/keycloak/data/import/prod-realm-export.json
-    command: start-dev --import-realm --http-port=8081 --proxy-headers xforwarded --hostname-strict=true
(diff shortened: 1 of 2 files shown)


thielpa (commented):
Not sure what we actually need, but maybe we can only enable the actual features we need instead of enabling all preview features. Not sure if there could be some security risks enabled with the other disabled features. 
See here for the docs: https://www.keycloak.org/server/features

TurkerKoc:
> https://www.keycloak.org/server/features

I agree. Looks like we only need to enable impersionation feature. [This one](https://www.keycloak.org/server/features#:~:text=Hostname%20Options%20V2-,impersonation,Ability%20for%20admins%20to%20impersonate%20users,-kerberos) I assume.

Btw enabling this feature has already some security risks :)

thielpa (approved):
Thanks :)

PR #549, opened 2025-03-17
chore: change data sync to run at 4 am every day instead of every hour

egekocabas:


TurkerKoc (approved):
LGTM! I think it's better to run data sync daily and at night. We can discuss this with the team as well. 🚀 

P.S: Maybe we can add data sync button into project settings to be able to run it manually. Also we can show the data sync status in the UI since it takes too much time

PR #571, opened 2025-03-18
feat: Truncate pipeline names + fix wrap in environment list

StefanNemeth:
### Motivation
The pipeline currently can look a bit crowded when the job names are long. Additionally, the deployment actions in the environment list is wrapped weirdly when there is a lot of info.

### Description
- Truncate the names to only a single line and show the full name on hover in the pipeline
- Group the deployment actions in a single div so it stays together and disable wrap for big screens

### Testing Instructions
Open a pipeline with a long enough job name and check the environment list with small and large screen sizes.

### Screenshots

**Pipeline:**
Screenshot 2025-03-18 at 17 01 50
image

**Environment list:**
Before:
image

After:
image

#### General

### DIFF
--- a/client/src/app/components/environments/environment-list/environment-list-view.component.html
+++ b/client/src/app/components/environments/environment-list/environment-list-view.component.html
@@ -55,18 +55,18 @@
                 <p-tag
                   severity="secondary"
                   rounded="true"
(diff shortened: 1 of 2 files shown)


egekocabas (approved):
Thank you for quickly fixing that bug 👍🏻

PR #585, opened 2025-03-21
feat: Environment name added to activity history

TurkerKoc:
mini-minor change again.
Added environment name into activity history page. 

image

### DIFF
--- a/client/src/app/pages/environment-deployment-history/environment-deployment-history.component.html
+++ b/client/src/app/pages/environment-deployment-history/environment-deployment-history.component.html
@@ -4,12 +4,13 @@
   </div>
   <span description> Here you can see all the locking, unlocking and deployment history. </span>
 </app-page-heading>
+<div class="text-gray-500 text-sm uppercase font-semibold tracking-wider mb-1" subheading>Environment: {{ environment()?.name }}</div>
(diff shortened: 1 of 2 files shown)


client/src/app/pages/environment-deployment-history/environment-deployment-history.component.html
```diff
@@ -4,12 +4,13 @@
   </div>
   <span description> Here you can see all the locking, unlocking and deployment history. </span>
 </app-page-heading>
+<div class="text-gray-500 text-sm uppercase font-semibold tracking-wider mb-1" subheading>Environment: {{ environment()?.name }}</div>
```
egekocabas, line 7: Should we use uppercase or use the environment name as is?

egekocabas (approved):
Just left one comment but it is not important, just a preference thing

other than that LGTM 👍🏻

client/src/app/pages/environment-deployment-history/environment-deployment-history.component.html
TurkerKoc: I would say uppercase is better since we are also using organization name/repo name in uppercase as well. 👍

PR #586, opened 2025-03-21
fix: Get branch on push event error

TurkerKoc:
We were getting errors on branch names containing "/"

reason for that was the way of extracting the branch name from the event. It was only getting the part before the first "/"


server/application-server/src/main/java/de/tum/cit/aet/helios/commit/github/GitHubPushMessageHandler.java
```diff
@@ -53,7 +53,17 @@ protected void handleInstalledRepositoryEvent(GHEventPayload.Push eventPayload)
       GHCommit commit = curRepo.getCommit(sha);
       commitSyncService.processCommit(commit, repository);
 
-      GHBranch branch = curRepo.getBranch(ref.split("/")[2]);
+      // Extract branch name by taking everything after the second slash
+      String[] refParts = ref.split("/");
```
thielpa, line 57: Cant we just use substring here?

Something like that (not tested): ref.substring(ref.indexOf("/") + 1);

thielpa (changes requested):

thielpa (approved):
Nice work!

PR #592, opened 2025-03-26
feat: First flaky test detection

TurkerKoc:
image
image
image
image


server/application-server/src/main/java/de/tum/cit/aet/helios/tests/TestResultService.java
```diff
@@ -122,19 +128,64 @@ private TestResultsDto getResultsFromRuns(
               .map(TestCase::getStatus);
         };
 
-    return new TestResultsDto(
-        testSuites.stream()
+    // Get repository ID from context
```
StefanNemeth, line 131: I think we can remove the getResultsFromRuns as we don't use it anymore so we don't need the duplicated logic

server/application-server/src/main/java/de/tum/cit/aet/helios/workflow/WorkflowRunRepository.java
```diff
@@ -104,14 +104,24 @@ Optional<String> findNthLatestCommitShaBehindHeadByBranchAndRepoId(
...
-          + "LEFT JOIN FETCH wr.testSuites "
           + "WHERE wr.headBranch = :branch "
           + "AND wr.headSha = :headSha "
           + "AND wr.repository.repositoryId = :repositoryId "
           + "AND wr.pullRequests IS EMPTY")
+  @EntityGraph(attributePaths = {"testSuites"})
   List<WorkflowRun> findByHeadBranchAndHeadShaAndRepositoryIdAndPullRequestsIsNullWithTestSuites(
       String branch, String headSha, Long repositoryId);
 
+  @Query(
```
StefanNemeth, line 115: I think we can remove this one as it's not used

server/application-server/src/main/resources/db/migration/V21__add_test_case_statistics.sql
```diff
@@ -0,0 +1,23 @@
+-- Add test case statistics table for flaky test detection
+CREATE TABLE test_case_statistics (
+    id BIGSERIAL PRIMARY KEY,
+    test_name VARCHAR(255) NOT NULL,
+    class_name VARCHAR(255) NOT NULL,
+    test_suite_name VARCHAR(255) NOT NULL,
+    branch_name VARCHAR(255) NOT NULL,
+    total_runs INTEGER NOT NULL DEFAULT 0,
+    failed_runs INTEGER NOT NULL DEFAULT 0,
+    failure_rate DOUBLE PRECISION NOT NULL DEFAULT 0.0,
```
StefanNemeth, line 10: I think we don't need to store the rate as it's always based on failed_runs and total_runs

server/application-server/src/main/resources/db/migration/V21__add_test_case_statistics.sql
```diff
@@ -0,0 +1,23 @@
...
+    is_flaky BOOLEAN NOT NULL DEFAULT FALSE,
+    last_updated TIMESTAMP WITH TIME ZONE NOT NULL,
+    
+    CONSTRAINT uk_test_case_statistics UNIQUE (test_name, class_name, test_suite_name, branch_name)
```
StefanNemeth, line 14: Maybe we don't need the test_suite_name here? But we need the repository id and some kind of test type identifier

StefanNemeth (commented):
Nice work! Left some comments :)

server/application-server/src/main/java/de/tum/cit/aet/helios/tests/TestCase.java
```diff
@@ -58,6 +59,21 @@ public void setClassName(String className) {
   @Column(name = "error_type")
   private String errorType;
```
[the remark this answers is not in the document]
TurkerKoc: After looking at it, transient fields actually used to temporarily store data in testcase before putting it into dto. It's better to keep it.

egekocabas (approved):
We have tested together and looks good, thanks 👍🏻 💯

PR #593, opened 2025-03-27
chore: remove Git LFS tracking

egekocabas:
### Motivation
GitHub imposes a 1 GB monthly bandwidth quota for Git LFS, and it turns out our build workflows — along with manual `git lfs pull` commands — have already consumed over 80% of that limit. With 15 days remaining until the quota resets, it's very likely we’ll hit the cap soon.

If we do, all `git lfs pull` operations will fail, meaning our CI builds won't be able to fetch required assets, and they’ll be missing from the final build artifacts. To avoid this situation, I created this issue to switch from Git LFS to regular Git tracking for these assets.

Also refer to the `#helios` slack channel where we talked this topic with Matthias.

image


egekocabas:
@ls1intum/helios are you comfortable with this change, any other opinion?

TurkerKoc (approved):
LGTM! Thanks for dealing with the issue before it's causing any problem 🚀

PR #599, opened 2025-03-28
refactor: apply header, sidebar and footer ui guidelines

thielpa:
image
image
image


egekocabas (approved):
Tested locally and works great 🚀 

This is just a personal remark, but the colors feel a bit strong to me. Maybe we could consider something softer? Though it might already be clear enough that we are in the readonly mode, so perhaps it’s fine as is, so feel free to disregard this comment.
image

PR #603, opened 2025-03-30
feat: show reviewers in helios client

gbanu:
Screenshot 2025-03-30 at 21 20 28

I wanted to make logic with the approval button conditional if the logged in user can review deployments. This would work for individual reviewers, but not for groups.

Unfortunately, we need extra permissions for our GitHub app to see if the user is a part of the team that is listed as a reviewer, therefore no specific logic is implemented for approvals
maybe we can discuss it with Stephan in the meeting how he wants to handle that


TurkerKoc (approved):
Thanks for implementing this feature so fast 🚀 💯 
Let's include this one also in our release!! Great Job!

PR #609, opened 2025-03-31
fix: Add real db pagination for test results

StefanNemeth:
### Motivation
Right now, the test result loading is extremely slow as we're not using pagination yet but loading all test cases into the server memory and sending it to the client.

### Description
Adds pagination to the client-side and the server-side. That involved a lot of refactoring to optimize queries for test statistics, previous tests, etc. There might still be some issues, but as the test results are not usable right now, I guess that's still better.

### Testing Instructions
You should test the functionalities of search, changing the page, showing only failed tests, showing updated/changed tests and testing the test statistics.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -13,6 +13,5 @@
   "editor.formatOnSave": true,
   "editor.codeActionsOnSave": {
     "source.fixAll": "always"
-  },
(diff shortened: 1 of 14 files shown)


thielpa (approved):
LGTM :robot: :confetti_ball:

PR #630, opened 2025-04-07
feat: Failure rates for all branches 

TurkerKoc:
**Failure Rate Updates:**
image
**Failure Rates and Flakiness Score:**
image
**Flakiness score calculation info pop-up:**
image

Removed isFlaky and FailureRate fields from TestcaseStatistics Entity.
Now we are only storing total runs and failed runs for default branch and all branches combined.
Then we are using this information to calculate failure rates and flakiness score.


gbanu (approved):
LGTM!
just left a couple of comments but, feel free to ignore

PR #633, opened 2025-04-08
fix: add delayed processing for GitHub `workflow_run` events

egekocabas:
### Motivation
We need to delay processing of GitHub `workflow_run` events by **10** (can be less or more) **seconds** so that the workflow has time to upload the `workflow-context` artifact. This artifact contains the `branch/SHA` information that triggered the workflow. Without waiting, we lose the ability to associate the run with the correct triggering branch/commit. We also need to ensure these events are processed sequentially, as our webhook processing is already sequential. 

### Description
- Added a single‐threaded TaskScheduler bean (workflowRunTaskScheduler) with poolSize(1) to guarantee tasks are handled one at a time.
- Refactored the workflow‐run handling to:
  - Schedule `workflow_run` events to run 10 seconds later
  - In `handleWorkflowRunEvent(...)`, we download the `workflow-context` artifact (if available) to extract the triggered workflow info (branch, SHA, etc.).
  - We call `workflowSyncService.processRunWithContext(...)` to store the extra fields (triggeredWorkflowRunId, headBranch, headSha), locate matching open PRs by branch/sha, and link them.
  - We finish by processing test results if the run meets the criterias
- Updated `GitHubWorkflowRunSyncService` with two methods: `processRun(...)` and `processRunWithContext(...)`, which delegate to a private `process(...)` method. This allows us to handle either the standard or “with context” scenario uniformly.

### Testing Instructions
Make sure that [your workflow saves](https://github.com/ls1intum/Artemis/blob/develop/.github/workflows/test-e2e.yml#L54-L65) the workflow-context file just like the Artemis E2E test workflow. Also triggering method of E2E test workflow must be 
```
on:
  workflow_run:
```

### DIFF
--- a/server/application-server/src/main/java/de/tum/cit/aet/helios/workflow/github/GitHubWorkflowRunMessageHandler.java
+++ b/server/application-server/src/main/java/de/tum/cit/aet/helios/workflow/github/GitHubWorkflowRunMessageHandler.java
@@ -13,7 +13,7 @@ import java.io.BufferedReader;
 import java.io.FilterInputStream;
 import java.io.IOException;
 import java.io.InputStreamReader;
-import java.util.Set;
+import java.time.Instant;
 import java.util.zip.ZipEntry;
 import java.util.zip.ZipInputStream;
 import lombok.RequiredArgsConstructor;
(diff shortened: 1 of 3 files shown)


server/application-server/src/main/java/de/tum/cit/aet/helios/workflow/github/GitHubWorkflowRunMessageHandler.java
```diff
@@ -36,6 +38,8 @@ public class GitHubWorkflowRunMessageHandler
   private final WorkflowRunService workflowRunService;
   private final WorkflowRunRepository workflowRunRepository;
   private final PullRequestRepository pullRequestRepository;
```
TurkerKoc, line 40: Looks like these 3 fields are not used anymore?

TurkerKoc (approved):
I've tested it in our test repo and everything works like charm. I also added on comment but we can ignore it. I will merge this one and let's see how it works in staging

PR #634, opened 2025-04-09
feat: Branches List View Filter by last 7 days and Expand/Collapse All

TurkerKoc:
Record for
Filtering Branches List View by 'Last 7 days'
Expand/Collapse all branches

P.S This should be tested with Artemis data since we are recursively traversing all the nodes, we should observe what happens with a lot of data.

https://github.com/user-attachments/assets/ace4d7c0-417f-442f-84a6-6dac297ef831

### DIFF
--- a/client/src/app/components/branches-table/branches-table.component.ts
+++ b/client/src/app/components/branches-table/branches-table.component.ts
@@ -48,6 +48,17 @@ const FILTER_OPTIONS = [
       }),
   },
   {
+    name: 'Last 7 Day',
+    filter: (branches: BranchInfoWithLink[]) =>
+      branches.filter(branch => {
+        const date = new Date(branch.updatedAt || '');
+        const staleThreshold = new Date();
+        staleThreshold.setDate(staleThreshold.getDate() - 7);
+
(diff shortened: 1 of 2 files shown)


client/src/app/components/branches-table/branches-table.component.ts
```diff
@@ -94,6 +105,8 @@ export class BranchTableComponent {
   searchTableService = inject(SearchTableService<BranchInfoWithLink>);
   keycloakService = inject(KeycloakService);
 
+  @ViewChild('table') treeTable!: TreeTable;
```
thielpa, line 108: Best practice: Please use the new version instead of the decorator viewChild('table');

thielpa (changes requested):
Good work, lgtm except for a minor code style 🙌

client/src/app/components/branches-table/branches-table.component.ts
TurkerKoc: Thanks for catching that! I wasn't aware it.
Converted it to signal-based approach 🙏

thielpa (approved):
LGTM!

PR #637, opened 2025-04-11
feat: Filter logs by log level

TurkerKoc:
- Implemented a filter with slider component to filter logs by level including the higher severity logs.
Log levels: OFF, FATAL, ERROR, WARN, INFO, DEBUG, TRACE, ALL
if you choose ERROR then FATAL and OFF will be included
if you choose DEBUG then INFO, WARN, ERROR, FATAL, OFF will be included

- I also added log level coloring to the lines
- Lines that doesn't contain any log level is kept in the text all the time
- Matching logic: It looks for lines that contains following pattern -> | <uppercase_log_level> |

Screen recording:

https://github.com/user-attachments/assets/4dcdb979-cbeb-4ebc-af97-f26de8cdfa1f


gbanu (approved):
I like the solution with the slider 👍 Nice work!

PR #639, opened 2025-04-11
feat: publish release with user token

gbanu:
Note: bases on generating release notes pr

Can be tested on repositories where the keycloak idp is installed in repository. (on development we use Helios-githubApp-local for authentication, so it can be tested only with ege-test-org repositories)


TurkerKoc:
One issue from the previous PR: 
- Generate release notes fails if there are no previous releases

Screenshot 2025-04-12 at 11 55 12

PR #644, opened 2025-04-12
fix: markdown and tabs issue

thielpa:
There were some issues with the markdown pipe, where the bg color was not set correctly for dark mode


gbanu (approved):
Crazy work, Paul! Kudos for dealing with the dark mode 👏 👏 👏

PR #646, opened 2025-04-13
fix: correct reassign logic for workflow_run in data sync

egekocabas:
### Motivation
The webhook-listener correctly handles workflow_run-typed workflow runs when they arrive via GitHub webhooks. However, during periodic data syncs (e.g., fallback polling or manual sync), the same logic failed to properly reassign workflow runs to their triggering workflows.


server/application-server/src/main/java/de/tum/cit/aet/helios/github/sync/GitHubDataSyncOrchestrator.java
```diff
@@ -415,10 +417,46 @@ public void syncRunsOfRepository(GHRepository repository, Optional<OffsetDateTim
       }
     }
 
-    workflowRuns.forEach(workflowRunSyncService::processRun);
+    workflowRuns.forEach(run -> {
+      try {
+
+        if (run.getEvent().equals(GHEvent.WORKFLOW_RUN)) {
+          log.info("Received workflow_run event, delaying processing");
```
TurkerKoc, line 424: I guess this is left from webhook handler, let's remove this log

TurkerKoc (approved):

PR #665, opened 2025-04-15
feat: sync environments button

gbanu:
it takes ~10 sec after the env creation before we can sync it

https://github.com/user-attachments/assets/e4b61c3a-7667-4058-b1b2-ba3d1140eee9


server/application-server/src/main/java/de/tum/cit/aet/helios/environment/EnvironmentService.java
```diff
@@ -631,4 +642,31 @@ public Optional<EnvironmentReviewersDto> getEnvironmentReviewers(Long environmen
...
+
+  /** Synchronizes the environments of the current repository with the GitHub repository. */
+  public void syncRepositoryEnvironments() {
+    Long repoId = RepositoryContext.getRepositoryId();
+    GitRepository repo = gitRepoRepository.findById(repoId).orElseThrow();
+    GHRepository ghRepository;
+    try {
+      ghRepository = this.gitHubService.getRepository(repo.getNameWithOwner());
+    } catch (IOException e) {
+      log.error("Failed to get GitHub repository: {}", e.getMessage());
```
thielpa, line 654: issue: Please add a throw statement here as currently it returns fetched successful when there is no repository found. You can use the EntityNotFound error here.

server/application-server/src/main/java/de/tum/cit/aet/helios/environment/EnvironmentService.java
```diff
@@ -631,4 +642,31 @@ public Optional<EnvironmentReviewersDto> getEnvironmentReviewers(Long environmen
...
+    try {
+      List<GitHubEnvironmentDto> gitHubEnvironmentDtoS =
+          gitHubService.getEnvironments(ghRepository);
+
+      for (GitHubEnvironmentDto gitHubEnvironmentDto : gitHubEnvironmentDtoS) {
+        environmentSyncService.processEnvironment(gitHubEnvironmentDto, ghRepository);
+      }
+
+    } catch (IOException e) {
+      log.error(
```
thielpa, line 666: Same here

thielpa (approved):
Nice and clean work here! 🥇 

I just found an edge case that should be adressed. Maybe we can also add a tooltip for the button that is saying that the sync is only working for envs older than the 10s, wdyt?

PR #669, opened 2025-04-18
fix: use tree shaking for icons

thielpa:


egekocabas (approved):
Thanks for getting rid of the old NgModule usage 🙏🏻  

I found couple missing icons but lets verify if we missed some in the staging 👍🏻

PR #672, opened 2025-04-20
feat: add version number to client

thielpa:


egekocabas (approved):
Thank you for this nice change, looks cool 💯 🚀  

I have just refactored build argument section, will deploy and lets see if that works on staging 👍🏻

PR #679, opened 2025-04-23
Fix/UI refactoring

egekocabas:
### Motivation
General UI fixes.

### Description
- Added missing `IconEyeOff` icon in `navigation-bar.component.ts`
- updated `pull-request-table.component.html` to be `flex-wrap` since too much label were creating horizontal scrollable (check it out the staging)
https://github.com/user-attachments/assets/f7ce10c6-2be0-4d91-9bdc-893c773aec21
- Sync Environments button were showing to non-loggedin users, covered with `isAtLeastMainainer`
- After setting the dark/light theme, if you refresh the page then it reverts back to the system default. Right now we persist the theme information in localstorage following the login in our [keycloak login theming code](https://github.com/ls1intum/Helios/blob/staging/keycloakify/index.html#L12-L43). In the local development, after you change your theme in helios would not reflect to the keycloak login page since the urls are different (helios-> localhost:4200, keycloak-> localhost:8081). But in the staging and production since the domains are the same, then setting the same `theme` using the `localstorage` would reflect to both of them.

### DIFF
--- a/client/src/app/components/environments/environment-list/environment-list-view.component.html
+++ b/client/src/app/components/environments/environment-list/environment-list-view.component.html
@@ -6,10 +6,12 @@
       <span class="hidden md:block">{{ isAdmin() ? 'Manage Environments' : 'Go to Environments' }}</span>
     </p-button>
   } @else {
-    <p-button label="" (click)="syncEnvironments()" pTooltip="May take ~10 seconds after environment creation on GitHub" tooltipPosition="top" class="mr-2">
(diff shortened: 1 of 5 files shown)


egekocabas:
image

Found out that in test table failed/skipped icons are missing and also the URLs for the workflowruns is missing

TurkerKoc (approved):
Thanks for adding all the icons and fixing minor problems. I also added 1-2 missing icons that I saw. I didin't want to create a new PR for 1-2 missing icons and included them in here.

PR #686, opened 2025-04-25
feat: Add configurable deployment workflow branch per environment

### LINKED ISSUE #649: Make branch that runs workflow for deployment configurable
## Context  
- For security reasons the runner for deploying to staging/prod only executes the staging deployment workflow file from the main or develop branches.
- Right now, when deploying a branch xy to staging the workflow is triggered for the xy branch. However, that won't work due to that limitation.
- Deployment of any branch/PR to staging should be possible, we don't want to limit that (requirement by @krusche ).

## Proposal
As suggested by @Hialus , we could make the branch, which we trigger the deployment workflow for, configurable for each environment. Meaning by default, they would trigger the workflow for the branch that will be deployed (current behaviour) and a maintainer could override it. So in this case, the staging environment would always trigger the workflow for the `main` or `develop` branch while still passing the branch that should be deployed as a variable.

StefanNemeth:
### Motivation
Fixes #649 

### Description
This PR adds the ability to configure which branch a deployment workflow should be triggered from on a per-environment basis. When set, workflow deployments will be triggered from the specified branch rather than the branch being deployed, while still using the correct deployment branch in the workflow parameters.

### Testing Instructions
1. Go into settings of an environment
2. Override the workflow branch of that environment to another one
3. Deploy to that environment and check whether the workflow is triggered for the branch you configured
4. Revert your change
5. Deploy to that environment and check whether the workflow is triggered for the branch you're deploying now

### Screenshots
image

#### General


server/application-server/src/main/java/de/tum/cit/aet/helios/environment/EnvironmentService.java
```diff
@@ -481,6 +481,7 @@ public Optional<EnvironmentDto> updateEnvironment(Long id, EnvironmentDto enviro
                 environment.setLockReservationExpiresAt(null);
               }
               environment.setEnabled(environmentDto.enabled());
+              environment.setDeploymentWorkflowBranch(environmentDto.deploymentWorkflowBranch());
```
TurkerKoc, line 484: It is possible to enter non-existing branch in here. Can we check whether branch exists or not before updating the branch informationf or environment? We can either check our db or make a request to github to check whether it exists or not. Second option is better in my opion

TurkerKoc (changes requested):
Thanks for adding this feature 🚀 
I tried it locally, it works pretty fine to dispatch workflow with the given branch. However, redeploying and unlocking is not working. The reason for this is not being able to match HeliosDeployment and Deployment with webhooks. Since branches are not matched HeliosDeployment object is not updated and we are not able to unlock or redeploy to that environment if we set another branch to start deployment. 

We have 2 options: 
- We can set heliosDeployment with updated branch while creating the HeliosDeployment object. --> [here](https://github.com/ls1intum/Helios/blob/01cb63bde2166170ec7253e76d6bd06c59ec40c7/server/application-server/src/main/java/de/tum/cit/aet/helios/deployment/DeploymentService.java#L141-L142) 
- Or we need to store triggering branch and compare it with that branch instead of delegated branch --> [Here where we are tyring to match HeliosDeployment](https://github.com/ls1intum/Helios/blob/01cb63bde2166170ec7253e76d6bd06c59ec40c7/server/application-server/src/main/java/de/tum/cit/aet/helios/deployment/github/GitHubDeploymentSyncService.java#L79-L81)


Recording for the errors:
- Even though deployment is finished it's not allowing users to redeploy or unlock
https://github.com/user-attachments/assets/034b55b0-0a62-41f7-b539-64a9b0961426

TurkerKoc (approved):
I tried it locally and it works perfectly! Only thing that we can consider later on is to store triggering branch and keep the history aligned with that one 🚀

PR #697, opened 2025-05-01
feat: notification page & server side changes

egekocabas:
#### Prerequisites:
- ...
#### Flow:
2. Navigate to Settings
3. ...


server/application-server/src/main/resources/db/migration/V34__notification.sql
```diff
@@ -0,0 +1,64 @@
+ALTER TABLE public."user"
+    ADD COLUMN has_logged_in         BOOLEAN NOT NULL DEFAULT FALSE,
+    ADD COLUMN notifications_enabled BOOLEAN NOT NULL DEFAULT FALSE,
+    ADD COLUMN notification_email    VARCHAR(255);
```
TurkerKoc, line 4: As we discussed, some users does not have any email in our database. We should fix data sync to get emails properly later on

server/application-server/src/main/resources/db/migration/V34__notification.sql
```diff
@@ -0,0 +1,64 @@
...
+UPDATE public."user"
+SET notification_email = email
+WHERE email IS NOT NULL;
+
+-- Create notification preference table
+CREATE TABLE public.notification_preference
+(
+    id      BIGSERIAL PRIMARY KEY,
+    user_id BIGINT       NOT NULL,
+    type    VARCHAR(255) NOT NULL,
```
TurkerKoc, line 16: If we want to add any other field than type we can create another table called 'notification_types'

client/src/app/pages/user-settings/user-settings.component.ts
```diff
@@ -0,0 +1,144 @@
...
+    ...updateUserSettingsMutation(),
+    onSuccess: () => {
+      this.messageService.add({
+        severity: 'success',
+        summary: 'Success',
+        detail: 'User settings updated successfully.',
+      });
+      this.queryClient.invalidateQueries({ queryKey: getUserSettingsQueryKey() });
+    },
+    onError: () => {
```
TurkerKoc, line 94: We have a global error handler but if this in is used for invalidateQueries we can also keep it

client/src/app/pages/user-settings/user-settings.component.html
```diff
@@ -0,0 +1,63 @@
+<div class="max-w-3xl mx-auto px-4 mt-8">
+  <p-card class="border border-surface-200 rounded-lg">
```
TurkerKoc, line 2: <p-card class="border border-surface-200 rounded-lg">
```suggestion
  <p-card class="rounded-lg">
```

TurkerKoc (approved):

client/src/app/pages/user-settings/user-settings.component.ts
egekocabas: I wanted to call the `invalidateQueries` after we got an error, lets leave it like this for not and we can check it out a way to maybe use the global one and then invalidate it 👍🏻

PR #714, opened 2025-05-15
feat: Cancelling deployment workflow run

TurkerKoc:
### Motivation

We want to give flexibility to users to directly cancel ongoing deployment using Helios.

### Description

This PR introduces following:
- Github Service to call cancel workflow run in Github
- Deployment Service to handle incoming cancellation request
- Deployment Controller to expose cancellation request to client
- If there is an ongoing deployment we show cancel button instead of deploy button in client side.

### Testing Instructions

- Basically starting a deployment and cancelling it afterwards

### Screencast
https://github.com/user-attachments/assets/0995fdd8-91e7-4775-973a-e410b85459e9

#### General
- [X] PR description explains the purpose and changes. [(f.e. following the 'Conventional Commits')](https://www.conventionalcommits.org/en/v1.0.0/)
- [X] Changes have been tested locally.
- [X] Screenshots have been attached.

#### Server
- [X] Code is performant and follows best practices
- [X] I documented the Java code using JavaDoc style.

#### Client
- [ ] I documented the TypeScript code using JSDoc style.
- [X] I added multiple screenshots/screencasts of my UI changes.
- [ ] I translated all newly inserted strings into English and German.


egekocabas (approved):
Love it 💯 

Only one thing maybe, in the client-side once we successfully send the cancel request, it takes time to GitHub to cancel the workflow and also there is a little time where we get the event & process it. So i was thinking maybe people shouldn't be able to cancel again (maybe within the first 10 seconds after their successful cancel request). Since they might think that i already clicked it but nothing happened, and can spam the button. We can also let it be like this, so no big preference from me, good job again 👍🏻

PR #741, opened 2025-05-21
feat: improve confirmation dialogs

egekocabas:
### Lock 
image
image

### Publish draft release
image
image

### Secret generation
image
image

##### Without required reviewers
image
image

##### With required reviewers
image
image

##### Without required reviewers
image
image

##### With required reviewers
image
image

##### Without required reviewers
image
image

##### With required reviewers
image
image

### DIFF
--- a/client/src/app/components/dialogs/deploy-confirmation/deploy-confirmation.component.html
+++ b/client/src/app/components/dialogs/deploy-confirmation/deploy-confirmation.component.html
@@ -17,7 +17,7 @@
       </div>
     </ng-container>
   } @else {
-    <div>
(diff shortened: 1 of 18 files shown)


thielpa:
Could you please add truncation to the modal headers when using environment names or something dynamic? :)

egekocabas:
> Could you please add truncation to the modal headers when using environment names or something dynamic? :)

I tried to truncate but it looks really bad and only very small environment names can fit into. I also don't want to increase the width of dialog, and also in order to be consistent with other popups; i just removed the environment name from the header. It is also written inside the dialog 😅  What do you think?

Issue #416, opened 2025-02-18
[Bug]:  Filtering PR for your PRs is shown even if you're not logged in


TurkerKoc:
### What happened?

- You still have an option to filter the PRs for your PRs even if you're not logged in

### Relevant log output

```shell

```

PR #750, opened 2025-05-23
fix: conditionally show user related PR filter

egekocabas:
### Motivation
Refer to https://github.com/ls1intum/Helios/issues/416

### Description
- If user logged in show user related filters, either show base ones
- Since we persist filter selection in the local storage, when loading the localstorage if any user related filters detected then do not set it and default to ALL prs


thielpa (approved):
Nice work! Looks good to me!

PR #763, opened 2025-05-28
feat: Adding comments to releases

TurkerKoc:
- Screencast
https://github.com/user-attachments/assets/a17f7597-5722-4f81-a8cc-2cefc2c1da39

Dialog update:
image


egekocabas (approved):
Thank you for this valuable feature 🙏🏻 

I have updated it so
- We can click the **same** evaluation button if we already did one. Since we introduced comments, user may forgot to add comment or later on want to add a comment.
- Also passed the existing evaluation's comment (if exist) to the popup (if it is the same evaluation). Thinking that user may want to update their existing comment.

PR #774, opened 2025-06-01
feat: workflow job status updates

gbanu:
- Saving no data regarding Job status in the backend.
- frontend -> backend -> github API
- from the video you can see that the status is updated with a lag, but eventually catches up
- job status is only shown during deployment to simplify the view

## Update:
- since api laggy anyways increase interval to 5 sec
- since api so laggy keep fetching for 1 min after deployment failed to display the results
- add waiting for approval
Screenshot 2025-06-01 at 18 26 21

- add fat button to GH if something failed
Screenshot 2025-06-01 at 19 46 41


egekocabas (approved):
LGTM 🙏🏻

PR #777, opened 2025-06-01
feat: skip repo-confirmation step for TEST environments

egekocabas:
### Motivation
Currently, every deployment requires the user to type the repository’s full “nameWithOwner” before the “Deploy” button becomes enabled. For `TEST` environments, we want to streamline the flow by removing this extra confirmation step and allowing an immediate deploy.

image
image


gbanu (approved):
LGTM! 🙏 thanks!

PR #835, opened 2025-12-05
fix: Implement scheduled recovery service to find stuck deployments and set their status to FAILED

meryemefe:
Fix: Implement scheduled recovery service to find stuck deployments and set their status to FAILED

### Motivation

There is an ongoing bug regarding to the deployments stuck in IN_PROGRESS state. The main root cause has not been found yet. As a workaround, a scheduled service is implemented to find such stuck deployments and set their status to FAILED.

### Description

There are two types of deployment class: Deployment and HeliosDeployment. Every hour, the scheduled recovery service finds both Deployment and HeliosDeployment whose statuses are IN_PROGRESS for more than an hour. Then, it updates these deployments' status to FAILED or FAILURE depending on the deployment class.

#### Prerequisites:
- The feature has been tested with unit tests and local tests by intentionally creating stuck deployments. However, in the server, it may not be easy to create such deployments intentionally. One way is to create a workflow that includes sleep command for an hour. Hence, this deployment will be IN_PROGRESS for more than an hour.
- Secondly, DeploymentRecoveryService will be running every hour. It can be monitored from the logs. 

#### General

#### Server

### DIFF
--- a/server/application-server/src/main/java/de/tum/cit/aet/helios/deployment/DeploymentRecoveryService.java
+++ b/server/application-server/src/main/java/de/tum/cit/aet/helios/deployment/DeploymentRecoveryService.java
@@ -83,7 +83,7 @@ public class DeploymentRecoveryService {
       log.warn(
           "Marking Helios deployment {} as FAILED, stuck in IN_PROGRESS state since {}",
           deployment.getId(),
-          deployment.getUpdatedAt());
+          deployment.getStatusUpdatedAt());
 
       deployment.setStatus(HeliosDeployment.Status.FAILED);
       heliosDeploymentRepository.save(deployment);


server/application-server/src/main/java/de/tum/cit/aet/helios/deployment/DeploymentRecoveryService.java
```diff
@@ -0,0 +1,94 @@
...
+              : deployment.getUpdatedAt());
+
+      deployment.setState(Deployment.State.FAILURE);
+      deploymentRepository.save(deployment);
+      updatedCount++;
+    }
+    return updatedCount;
+  }
+
+  private int processStuckHeliosDeployments(OffsetDateTime timeThreshold) {
```
mertilginoglu: What about merging this with `processStuckDeployments`? They seem mostly the same

PR #850, opened 2025-12-21
feat: enhance pull request synchronization to include closed PRs and support optional cutoff date

mertilginoglu:
### Motivation

  Helios can show stale PR states when webhook events are missed (e.g., NATS downtime). The scheduled sync only refreshed open PRs, so merged/closed PRs remained “open” in Helios. This change ensures periodic sync also updates closed/merged PRs within the configured timeframe.

  ### Description

  - Sync pull requests using state=ALL, bounded by the monitoring timeframe cutoff.
  - Added unit tests covering PR sync behavior with and without the cutoff.

  ### Testing Instructions
This feature runs during the periodic sync.  

  ### Checklist

  #### General

  - [x] PR description explains the purpose and changes.

  #### Server

### DIFF
--- a/server/application-server/src/main/java/de/tum/cit/aet/helios/github/sync/GitHubDataSyncService.java
+++ b/server/application-server/src/main/java/de/tum/cit/aet/helios/github/sync/GitHubDataSyncService.java
@@ -41,6 +41,8 @@ public class GitHubDataSyncService {
 
   public void syncRepositoryData(String repositoryNameWithOwner) {
     var cutoffDate = OffsetDateTime.now().minusDays(timeframe);
+    // Keep a broader PR window to recover from missed webhook events.
+    var pullRequestCutoffDate = cutoffDate;
 
     // Get last sync time
     var lastSync =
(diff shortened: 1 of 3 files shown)


server/application-server/src/main/java/de/tum/cit/aet/helios/github/sync/GitHubDataSyncService.java
```diff
@@ -116,8 +118,10 @@ public void syncRepositoryData(String repositoryNameWithOwner) {
       logSeparator();
       var step3Start = Instant.now();
       log.info(
-          "Repository: {} --> [Step 3/10] Syncing Open Pull Requests...", repositoryNameWithOwner);
-      dataSyncOrchestrator.syncPullRequestsOfRepository(ghRepository);
+          "Repository: {} --> [Step 3/10] Syncing Pull Requests (Open + Closed)...",
+          repositoryNameWithOwner);
+      dataSyncOrchestrator.syncPullRequestsOfRepository(
+          ghRepository, Optional.of(pullRequestCutoffDate));
```
meryemefe, line 124: Hi Mert, thank you for your PR to fix the important inconsistency caused by missed webhook events. Actually, webhook failures are expected to be rare, and under normal operating conditions we should not see crashes or missed events frequently. My concern is that when the system is functioning without failure, synchronizing pull requests for the last 7 days (or whatever the monitoring.timeframe is) may introduce unnecessary load and complexity. 

I suggest creating an endpoint that allows us to manually trigger a full-synchronization starting from a specified datetime. This would give us control to run the broader sync only when needed. What do you think?

Issue #866, opened 2026-01-11
[Bug]: Value too long error for type character varying(255) in workflow_run


meryemefe:
### Contact Details

@meryemefe 

### What happened?

In some workflow runs or pull requests the name or displayed title can be longer than 255 character. It is a rare case. It is observed that usually bot-generated PRs have a longer title. Eg.: 
https://github.com/ls1intum/edutelligence/actions/runs/21819018087

In such case, the PR or workflow run are trying to be processed by Helios. However, because these fields are varchar (255) in the database, it throws exception. 

### Steps To Reproduce

1. Create a pull request with a name longer than 255 character.
2. Check the Helios logs.

### Relevant log output

```shell
2026-01-10T06:23:28.007Z DEBUG 1 --- [Helios] [default-scheduler-1] t.c.a.h.w.g.GitHubWorkflowRunSyncService : Workflow run 20709432328 is not a deployment workflow run
2026-01-10T06:23:28.023Z  WARN 1 --- [Helios] [default-scheduler-1] o.h.engine.jdbc.spi.SqlExceptionHelper   : SQL Error: 0, SQLState: 22001
2026-01-10T06:23:28.023Z ERROR 1 --- [Helios] [default-scheduler-1] o.h.engine.jdbc.spi.SqlExceptionHelper   : ERROR: value too long for type character varying(255)
[...]

PR #880, opened 2026-02-15
fix: truncate long titles and names in GitHub issue and workflow run converters

meryemefe:
fix: truncate long titles and names in GitHub issue and workflow run converters

### Motivation

Some PRs or workflow runs can have titles and names longer than 255 characters. When syncing these via the GitHub API, Helios tries to save them into the database. However, PostgreSQL raises: `ERROR: value too long for type character varying(255)`

This PR fixes issue #866 .  

### Description

This issue causes sync failures for affected workflow runs (and similarly for issues/PRs with long titles). This change truncates long titles and names to 255 characters in the converters with a warning log. Since long titles/names are actually unusual, I don't prefer changing database schema and increasing the character limit, which can lead to the poor data modeling.

#### Flow:
1. First, in your local environment, create a PR with a name longer than 255 character and see the error logs containing `[ERROR: value too long for type character varying(255)]`
2. Then, checkout to this branch and create another PR.
3. In the logs, you should see a warning log containing `Truncated ... field '...' from .. to 255 characters for source ...: ...`
4. Verify than the PR is successfully added to your database.
5. Do the same steps for a workflow.

#### General

#### Server

### DIFF
--- a/server/application-server/src/main/java/de/tum/cit/aet/helios/issue/github/GitHubIssueConverter.java
+++ b/server/application-server/src/main/java/de/tum/cit/aet/helios/issue/github/GitHubIssueConverter.java
@@ -50,7 +50,7 @@ public class GitHubIssueConverter extends BaseGitServiceEntityConverter<GHIssue,
       return value;
     }
     String truncatedValue = value.substring(0, maxLength);
-    log.warn(
+    log.info(
         "Truncated issue field '{}' from {} to {} characters for source {}: '{}'",
         fieldName,
         value.length(),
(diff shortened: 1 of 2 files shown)


server/application-server/src/main/java/de/tum/cit/aet/helios/issue/github/GitHubIssueConverter.java
```diff
@@ -40,4 +44,19 @@ private Issue.State convertState(GHIssueState state) {
         return Issue.State.CLOSED;
     }
   }
+
+  private static String truncate(String value, int maxLength, String fieldName, long sourceId) {
+    if (value == null || value.length() <= maxLength) {
+      return value;
+    }
+    String truncatedValue = value.substring(0, maxLength);
+    log.warn(
```
mertilginoglu: Maybe we change this to info?

server/application-server/src/main/java/de/tum/cit/aet/helios/workflow/github/GitHubWorkflowRunConverter.java
```diff
@@ -61,6 +66,21 @@ public WorkflowRun update(@NonNull GHWorkflowRun source, @NonNull WorkflowRun wo
     return workflowRun;
   }
 
+  private static String truncate(String value, int maxLength, String fieldName, long sourceId) {
+    if (value == null || value.length() <= maxLength) {
+      return value;
+    }
+    String truncatedValue = value.substring(0, maxLength);
+    log.warn(
```
mertilginoglu: This too

mertilginoglu (approved):

PR #900, opened 2026-03-08
fix: add admin pull request state reconciliation endpoint

mertilginoglu:
### Motivation

  Open pull requests can remain stale in Helios when GitHub state changes are missed or delayed. This makes the local PR state unreliable for reviewers and administrators and makes it harder to recover from synchronization gaps.

  ### Description

  - add an admin-only pull request state reconciliation endpoint for a repository, including optional dryRun support
  - add a reconciliation service and result DTO that:
      - loads locally open pull requests for a repository
      - checks their current state on GitHub
      - reprocesses pull requests that are already closed on GitHub
      - reports updated, unchanged, missing, and failed reconciliations
  - extend error handling so authorization failures on the new endpoint are returned as 403 Forbidden
  - add focused tests for:
      - the reconciliation service behavior
      - controller authorization and dryRun handling
      - pull request sync cutoff behavior in the GitHub data sync orchestrator
  - include minor supporting cleanup in related server files

  ### Testing Instructions

  #### Prerequisites:

  - local server environment for Helios
  - a repository connected to Helios
  - one admin user and one non-admin user
  - a repository ID with pull requests available in Helios

  #### Flow:

  1. Run the targeted server tests for the new reconciliation flow and controller coverage.
  2. Start the application server locally.
  3. As a non-admin user, call POST `/api/pullrequests/repository/{repositoryId}/reconcile-state` and verify the request is rejected with 403 Forbidden.
  4. As an admin user, call POST `/api/pullrequests/repository/{repositoryId}/reconcile-state?dryRun=true` and verify the response contains reconciliation statistics without modifying data.
  5. Call POST `/api/pullrequests/repository/{repositoryId}/reconcile-state` as an admin user and verify that pull requests which are locally open but already closed on GitHub are reconciled correctly.
  6. Verify that missing GitHub pull requests are counted as missing and unexpected GitHub/API failures are reported as errors in the response.

  ### Checklist

  #### General

  - [x] PR description explains the purpose and changes. (f.e. following the 'Conventional Commits') (https://www.conventionalcommits.org/en/v1.0.0/)

  #### Server

### DIFF
--- a/server/application-server/src/main/java/de/tum/cit/aet/helios/pullrequest/PullRequestStateReconciliationService.java
+++ b/server/application-server/src/main/java/de/tum/cit/aet/helios/pullrequest/PullRequestStateReconciliationService.java
@@ -1,0 +1,127 @@
+package de.tum.cit.aet.helios.pullrequest;
+
+import de.tum.cit.aet.helios.github.GitHubFacade;
+import de.tum.cit.aet.helios.gitrepo.GitRepoRepository;
+import de.tum.cit.aet.helios.gitrepo.GitRepository;
+import de.tum.cit.aet.helios.issue.Issue;
+import de.tum.cit.aet.helios.pullrequest.github.GitHubPullRequestSyncService;
+import jakarta.persistence.EntityNotFoundException;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.regex.Pattern;
+import lombok.RequiredArgsConstructor;
+import lombok.extern.log4j.Log4j2;
+import org.kohsuke.github.GHIssueState;
+import org.springframework.stereotype.Service;
+import org.springframework.transaction.annotation.Transactional;
+
+@Service
+@Log4j2
+@RequiredArgsConstructor
+public class PullRequestStateReconciliationService {
+
+  private static final Pattern NOT_FOUND_PATTERN =
+      Pattern.compile("not found|\"status\"\\s*:\\s*\"404\"", Pattern.CASE_INSENSITIVE);
+
+  private final PullRequestRepository pullRequestRepository;
+  private final GitRepoRepository gitRepoRepository;
+  private final GitHubPullRequestSyncService pullRequestSyncService;
+  private final GitHubFacade github;
+
+  @Transactional
+  public PullRequestStateReconciliationResultDto reconcilePullRequestState(
+      Long repositoryId, boolean dryRun)
+      throws IOException {
+    GitRepository repository =
+        gitRepoRepository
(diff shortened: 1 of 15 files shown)


server/application-server/src/main/java/de/tum/cit/aet/helios/pullrequest/PullRequestStateReconciliationService.java
```diff
@@ -0,0 +1,127 @@
...
+            log.info(
+                "Reconciled local pull request {} (#{}) to closed state for repository {}.",
+                pullRequest.getId(),
+                pullRequest.getNumber(),
+                repository.getNameWithOwner());
+          }
+        } else {
+          unchangedCount++;
+        }
+      } catch (IOException e) {
```
meryemefe: I have the same comment with codex-connector :) Maybe we can catch the Exception instead of only IOException. Also, just a suggestion: we can extract this logic inside `for` loop to a private `reconcilePullRequest` method for a single pull request. It can increase the readibility.

server/application-server/src/main/java/de/tum/cit/aet/helios/pullrequest/PullRequestStateReconciliationService.java
```diff
@@ -0,0 +1,127 @@
...
+
+  private static final Pattern NOT_FOUND_PATTERN =
+      Pattern.compile("not found|\"status\"\\s*:\\s*\"404\"", Pattern.CASE_INSENSITIVE);
+
+  private final PullRequestRepository pullRequestRepository;
+  private final GitRepoRepository gitRepoRepository;
+  private final GitHubPullRequestSyncService pullRequestSyncService;
+  private final GitHubFacade github;
+
+  @Transactional
```
meryemefe: I am not sure about this `@Transactional`. `processPullRequest` method is already Transactional. I think each pull request update should be atomic independently.

meryemefe:
I have a general comment: In local tests, I got 403 first. Then, I realized that admin authorization requires X-Repository-Id header, so we will get 403 without it even if we have admin access rights. It would be good to document it for next developers. Maybe we can also think about an Admin Dashboard for such operations.

mertilginoglu:
> I have a general comment: In local tests, I got 403 first. Then, I realized that admin authorization requires X-Repository-Id header, so we will get 403 without it even if we have admin access rights. It would be good to document it for next developers. Maybe we can also think about an Admin Dashboard for such operations.

I added a new service to check repository related authorization. I noticed that in previous case admins in different repos can run this endpoint also for other repositories.

meryemefe (approved):
I tested it, LGTM

PR #903, opened 2026-03-08
fix: update repository overview pagination

### LINKED ISSUE #710: Unexpected pagination behavior (repository overview)
### Contact Details

Artemis DM

### What happened?

Imo, showing 6 repositories per page is a bit unexpected. I know 6 repositories fit well on e.g. an iPad screen, but with a larger or smaller screen this is not the case. Also, compared to other applications, 6 is quite unusal for pagination.
Therefore, I would increase the number of repositories per page or add the option to manually adapt it (see https://primeng.org/paginator). A custom number depending on the screen size would be the most elegant solution.

This is the according line:
https://github.com/ls1intum/Helios/blob/3d3626a0ba4a6a4bb7ff15bac0c56fc2b8363fc5/client/src/app/pages/repository-overview/repository-overview.component.html#L43

### Steps To Reproduce

1. Use a large monitor with reasonable size settings
2. Open Helios home page
3. Perceive unexpected pagination behavior:

Image
Here, 2 repositories are shown on page 2, even though they would fit perfectly.

_No response_

### Relevant log output

```shell

```

mertilginoglu:
### Motivation

Fixes #710.

The repository overview currently paginates with a fixed page size of 6 repositories, which feels unusually low on larger screens and often leaves the second page half empty even when more cards would fit comfortably. This change makes the pagination behavior more practical and gives users control over how many repositories are shown per page.

### Description

- increase the default repository overview page size from 6 to 12
- add page size options so users can switch between 6, 12, 24, and 48 repositories per page
- keep the selected paginator state in the component and apply it to both the loaded and loading states
- update the loading skeletons to match the active page size for a more consistent UI
- add client tests covering the default paginator configuration, loading placeholder count, and page change handling

#### Prerequisites:

- Helios client dependencies are installed
- A Helios instance with enough connected repositories to span multiple pages

#### Flow:

1. Start the client and log in to Helios as a Developer.
2. Open the repository overview page.
3. Verify that the initial page shows 12 repositories instead of 6.
4. Change the paginator page size and verify that the available options are 6, 12, 24, and 48.
5. Navigate between pages and confirm that the selected page size is applied correctly.
6. Verify that the pagination no longer leaves unexpectedly sparse pages when the repository list would fit on the current screen more naturally.
7. Run the repository overview tests and confirm the paginator-related test cases pass.

### Screenshots

Screenshot 2026-03-14 at 16 33 30

#### General

#### Client

### DIFF
--- a/client/src/app/components/table-filter/table-filter.component.html
+++ b/client/src/app/components/table-filter/table-filter.component.html
@@ -8,15 +8,7 @@
       }
     </p-button>
   }
-  <input
-    pInputText
-    [(ngModel)]="searchTableService().searchValue"
-    type="text"
-    pSize="small"
-    placeholder="Search"
-    class="w-full"
-    (input)="searchTableService().onInput(table(), $event)"
-  />
+  <input pInputText [(ngModel)]="searchTableService().searchValue" type="text" pSize="small" placeholder="Search" class="w-full" (input)="onInput($event)" />
 </div>
 
 <p-popover #op>
--- a/client/src/app/components/table-filter/table-filter.component.ts
+++ b/client/src/app/components/table-filter/table-filter.component.ts
@@ -1,4 +1,4 @@
-import { Component, computed, input, viewChild } from '@angular/core';
+import { Component, computed, input, output, viewChild } from '@angular/core';
 import { FormsModule } from '@angular/forms';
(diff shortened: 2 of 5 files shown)


meryemefe (approved):
I tested in my local and reviewed the changes. LGTM

PR #912, opened 2026-03-13
feat: add workflow runs by repository endpoint with pagination and filtering options

meryemefe:
### Motivation

This feature adds a paginated, filterable Workflow Runs listing under the CI/CD menu so users can inspect workflow executions across a repository, ordered by start time and with metadata like status and test processing status. This provides a centralized overview of CI activity beyond individual PRs/branches. #913

### Description

Since the feature contains a lot of changes, this PR includes only backend changes.

- Added `GET /api/workflows/runs` endpoint in `WorkflowRunController `returning a `PaginatedWorkflowRunsResponse` with:
  - `runs: List<WorkflowRunDto>`
  - `page, size, totalElements, totalPages`.
- Extended `WorkflowRunService` with `getPaginatedWorkflowRuns(WorkflowRunPageRequest)` that:
  - Scopes results to the current RepositoryContext repository.
  - Applies server-side search over name, displayTitle, headBranch, and headSha.
  - Applies high-level filters on status/conclusion via `WorkflowRunFilterType`.
  - Sorts by `runStartedAt DESC`, `createdAt DESC` by default, or by a requested field (`name, status, branch, runStartedAt, updatedAt`) with direction.
- Extended existing `WorkflowRunDto` to be suitable for list views:
  - Added testProcessingStatus, headBranch, headSha, runStartedAt, createdAt, updatedAt.
- Added `PaginatedWorkflowRunsResponse` record in workflow.pagination.
- Made `WorkflowRunRepository` extend `JpaSpecificationExecutor<WorkflowRun>` to support specifications and pageable queries (`findAll(Specification<WorkflowRun>, Pageable)`).

#### Prerequisites:
- Helios running with a repository that has existing workflow runs.
- A user with access to that repository.
#### Flow:
1. Manual API verification
1.1. Start the application server.
1.2. Call GET `/api/workflows/runs` for a repository with a configured RepositoryContext.
1.3. Verify response shape:
runs array of WorkflowRunDto objects.
page = 1, size = 20, totalElements, totalPages set appropriately.
1.4. Confirm that runs are ordered by runStartedAt (then createdAt) descending.
1.5. Repeat the API calls with different `page, size, sortField, sortDirection, searchTerm` parameters.
2. Automated Tests: Open server directory and run the test cases by using `./gradlew :application-server:test` command.

#### General

#### Server

### DIFF
--- a/server/application-server/src/main/java/de/tum/cit/aet/helios/workflow/pagination/WorkflowRunPageRequest.java
+++ b/server/application-server/src/main/java/de/tum/cit/aet/helios/workflow/pagination/WorkflowRunPageRequest.java
@@ -1,0 +1,17 @@
+package de.tum.cit.aet.helios.workflow.pagination;
+
+import lombok.Builder;
+import lombok.Getter;
+import lombok.Setter;
+
+@Getter
+@Setter
+@Builder
+public class WorkflowRunPageRequest {
+  private int page = 1;
+  private int size = 20;
+  private String sortField;
+  private String sortDirection;
+  private WorkflowRunFilterType filterType = WorkflowRunFilterType.ALL;
+  private String searchTerm;
+}
--- a/server/application-server/src/main/java/de/tum/cit/aet/helios/workflow/pagination/WorkflowRunPageRequest.java
+++ b/server/application-server/src/main/java/de/tum/cit/aet/helios/workflow/pagination/WorkflowRunPageRequest.java
@@ -8,10 +8,13 @@ import lombok.Setter;
 @Setter
 @Builder
 public class WorkflowRunPageRequest {
+  @Builder.Default
   private int page = 1;
+  @Builder.Default
   private int size = 20;
   private String sortField;
   private String sortDirection;
+  @Builder.Default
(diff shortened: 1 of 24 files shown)


server/application-server/src/main/java/de/tum/cit/aet/helios/workflow/pagination/WorkflowRunPageRequest.java
```diff
@@ -0,0 +1,17 @@
...
+
+import lombok.Builder;
+import lombok.Getter;
+import lombok.Setter;
+
+@Getter
+@Setter
+@Builder
+public class WorkflowRunPageRequest {
+  private int page = 1;
```
mertilginoglu, line 12: We provide default values in the controller but if we want to use this builder somewhere else we would need `Builder.Default` to have default values.

meryemefe: Thank you for your comment. I added @Builder.Default

mertilginoglu (approved):

Issue #913, opened 2026-03-13
[Feature]: Workflow Runs Tab


meryemefe:
### Describe the feature

We will implement an workflow runs tab in CI/CD menu.

PR #916, opened 2026-03-14
feat: implement workflow runs menu and table component

meryemefe:
### Motivation

This feature adds a paginated, filterable Workflow Runs listing under the CI/CD menu so users can inspect workflow executions across a repository, ordered by start time and with metadata like status and test processing status. This provides a centralized overview of CI activity beyond individual PRs/branches. #913

### Description

Workflow Run tab is added to CI/CD menu. Workflow runs can be view in timestamp descending order with 
name, status, test processing status, started at, updated at, branch and sha fields.

#### Prerequisites:
- GitHub repository connected to Helios with workflows configured.
- At least a few workflow runs existing in the repository.
- A user account that can log into Helios as a Developer.
#### Flow:
2. Navigate to the target Repository.
3. Open the CI/CD section in the sidebar.
4. Ensure you land on the Workflow Runs tab.
5. Confirm a paginated table of workflow runs is displayed. 
6. Use the filter dropdown and verify the table updates as expected.
7. Change page size and navigate between pages to ensure server‑side pagination continues to work.
8. Click the GitHub button in a row and verify it opens the workflow run in GitHub in a new tab.

### Screenshots

Screenshot 2026-03-14 at 10 48 42
Screenshot 2026-03-14 at 10 49 04
Screenshot 2026-03-14 at 10 49 22
Screenshot 2026-03-14 at 10 51 03

#### General

#### Server

#### Client

### DIFF
--- a/client/src/app/app.routes.ts
+++ b/client/src/app/app.routes.ts
@@ -80,7 +80,7 @@ export const routes: Routes = [
             children: [
               {
                 path: '',
-                redirectTo: 'runs',
+                redirectTo: 'pr',
                 pathMatch: 'full',
               },
               {
--- a/client/src/app/app.routes.ts
+++ b/client/src/app/app.routes.ts
@@ -80,8 +80,12 @@ export const routes: Routes = [
             children: [
               {
                 path: '',
-                redirectTo: 'pr',
+                redirectTo: 'runs',
                 pathMatch: 'full',
+              },
+              {
+                path: 'runs',
+                loadComponent: () => import('@app/pages/workflow-run-list/workflow-run-list.component').then(m => m.WorkflowRunListComponent),
               },
               {
                 path: 'pr',
--- a/client/src/app/components/workflow-runs-table/workflow-runs-table.component.html
+++ b/client/src/app/components/workflow-runs-table/workflow-runs-table.component.html
@@ -1,0 +1,103 @@
+<p-table
+  #table
+  [rowHover]="true"
+  [value]="runs()"
+  [lazy]="true"
+  [totalRecords]="totalElements()"
+  [paginator]="true"
+  [rows]="paginationService.size()"
+  [first]="(paginationService.page() - 1) * paginationService.size()"
+  [rowsPerPageOptions]="[5, 10, 20, 30, 50]"
+  [loading]="query.isPending()"
+  (onPage)="onPage($event)"
+  (sortFunction)="onSort($event)"
+  [customSort]="true"
+  [sortField]="paginationService.sortField()"
+  [sortOrder]="paginationService.sortDirection() === 'asc' ? 1 : -1"
+>
+  <ng-template #header>
+    <tr>
+      <th>
+        <app-table-filter-paginated [paginationService]="typedPaginationService" [table]="table" />
+      </th>
+      <th class="hidden md:table-cell">Status</th>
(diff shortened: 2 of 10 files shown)


client/src/app/components/workflow-runs-table/workflow-runs-table.component.html
```diff
@@ -0,0 +1,103 @@
...
+        </td>
+      </tr>
+    }
+  </ng-template>
+
+  <ng-template pTemplate="emptymessage">
+    <tr>
+      <td colspan="5">
+        <div class="flex flex-col gap-2 p-20 justify-center items-center">
+          <i-tabler name="git-pull-request" class="!h-20 !w-20 text-red-500" />
```
mertilginoglu: Do we want to have `git-pull-request` related icon for empty scenario?

client/src/app/app.routes.ts
```diff
@@ -80,9 +80,13 @@ export const routes: Routes = [
             children: [
               {
                 path: '',
-                redirectTo: 'pr',
```
mertilginoglu, line 83: Not sure if we should default to workflow page. I think keeping PR page as default could be more useful for our users.

client/src/app/components/workflow-runs-table/workflow-runs-table.component.ts
```diff
@@ -0,0 +1,203 @@
...
+        sortDirection: paginationState.sortDirection,
+        filterType: filterType,
+        searchTerm: paginationState.searchTerm,
+      },
+    });
+  });
+
+  query = injectQuery(() => this.queryOptions());
+
+  constructor() {
```
mertilginoglu: Got this review from Claude. Also tested in my local and it seems like we are double fetching when we change the filter.


Double-fetch on every pagination/filter change
  The `constructor` effect watches `paginationService.paginationState()` and calls `this.query.refetch()`. But `injectQuery(() => this.queryOptions())` already re-fetches automatically
  when `queryOptions` (a computed depending on `paginationState`) changes. Every state change triggers two fetches. The effect in the constructor should be removed entirely. The
  explicit `this.query.refetch()` call in onPage has the same issue.

```
  // Remove this entirely from constructor:
  effect(() => {
    this.paginationService.paginationState();
    if (this.query.data()) {
      this.query.refetch();
    }
  });
```

```
  // And remove this from onPage:
  this.query.refetch();
```

client/src/app/components/workflow-runs-table/workflow-runs-table.component.html
meryemefe: Yes thank you, I left it as it is to check it later, and forgot.. Player icon could be better.

client/src/app/app.routes.ts
meryemefe: Actually I am also not sure, but let me change it PR again. If users prefer to see Runs first, we can change it later

client/src/app/components/workflow-runs-table/workflow-runs-table.component.ts
meryemefe: Thank you, that's very important comment. I think we have the same problem in pull requests page too because I used that page as a reference. But, this PR is already big, let me fix Pull Request page in another PR.

mertilginoglu (approved):

PR #919, opened 2026-03-18
fix: simplify deploy confirmation dialog messaging

mertilginoglu:
Fixes model dialog
Screenshot 2026-03-18 at 10 38 25


meryemefe (approved):
I tested locally, LGTM

PR #939, opened 2026-03-29
feat: use configured deploy job for deployment time estimates

mertilginoglu:
### Motivation

Helios currently uses fixed timing assumptions when estimating how long a deployment will stay in pending and in-progress states. That makes the deployment progress view inaccurate for workflows whose actual deploy step starts much later or earlier than the default assumption. This change lets repositories configure the GitHub Actions job that represents the real deployment start so Helios can calculate more accurate build and deploy duration estimates.

### Description

- add persistent deployment workflow configuration for GitHub workflows, including a stored `deployJobName` per workflow
- expose new settings endpoints and generated OpenAPI client methods to read and update deployment workflow configuration
- extend repository settings in the client with a per-workflow "Deployment Job" configuration dialog
- track deployment job timing from GitHub workflow job events and use it to split estimated pre-deployment time from actual deployment time
- store and expose estimated build and deploy durations on deployment/environment DTOs so the client can render more accurate progress timing
- update deployment timing logic in the client to use backend-provided estimates instead of relying only on static defaults
- add migration, service, and test updates needed for the new workflow timing and deployment configuration flow

#### Prerequisites:
- Local Helios setup with server and client running
- A repository connected to Helios that has GitHub Actions workflows and at least one environment using a deployment workflow
- A user with maintainer access for repository settings in Helios
- A GitHub Account without having any additional access-rights (e.g. admin, owner)

#### Flow:
1. Log in to Helios with a maintainer account
2. Open a repository that has deployment workflows configured for at least one environment
3. Navigate to `Settings`
4. In the workflow table, click `Configure` in the new `Deployment Job` column for a deployment workflow
5. Enter the GitHub Actions job name that represents the actual deployment step and save it
6. Reopen the dialog and verify the configured job name is persisted
7. Trigger a deployment for an environment that uses this workflow
8. After GitHub workflow job data has been synced back to Helios, open the environment or deployment progress view and verify the timing estimate reflects the configured deploy job instead of only the old fixed defaults
9. Repeat the flow with a workflow that has no deployment job configuration and verify Helios still falls back to the default estimate behavior

#### General

#### Server


server/application-server/src/main/java/de/tum/cit/aet/helios/environment/EnvironmentService.java
```diff
@@ -122,6 +127,21 @@ public List<EnvironmentDto> getEnvironmentsByRepositoryId(Long repositoryId) {
    * @return A wrapper object containing the latest Deployment or HeliosDeployment, or an empty
    *     result if no deployments exist.
    */
+  private DeploymentDurationEstimate computeEstimate(Environment environment) {
```
meryemefe, line 127: It could be better if we replace this `computeEstimate` method after `findLatestDeployment` method because now it has wrong JavaDoc

server/application-server/src/main/java/de/tum/cit/aet/helios/tests/TestResultProcessor.java
```diff
@@ -86,7 +86,9 @@ public void processRun(WorkflowRun workflowRun) {
       updateTestStatisticsIfDefaultBranch(testSuites, workflowRun);
       // HOTFIX: Flakiness recomputation is temporarily disabled to avoid OOMs in production.
       // TODO: Re-enable after memory-safe flakiness path is deployed.
-      log.warn("HOTFIX active: skipping flakiness update for workflow run {}", workflowRun.getName());
+      log.warn(
```
meryemefe: I enabled this flakiness scores update method in staging, which causes merge-conflict now.

meryemefe (commented):
I tested locally. It works very well. I have only some minor comments. Also, I would like to remind that you need to update the version of flyway while merging conflicts.

PR #983, opened 2026-04-27
feat: implement workflow run cancellation synchronization and add HeliosDeploymentWorkflowRunSyncService

### LINKED ISSUE #990: [Bug]:  Cancellation Issue for Completed Deployments
### Contact Details

_No response_

### What happened?

Sometimes, even though deployments has been completed in GitHub, processing the respective webhook event can take a while. In this case, users don't see the deployment as completed in the UI and they face issues when they try to cancel the deployment.

### Steps To Reproduce

_No response_

_No response_

### Relevant log output

```shell

```

mertilginoglu:
### Motivation
GitHub can reject cancellation when a workflow run has already reached a terminal state, leaving Helios deployments stale even though GitHub has the final status. Queued workflow runs also need to stay visible as a distinct waiting state so reviewers understand that deployment is blocked on GitHub Actions runner availability.

Fixes #990 

### Description
- Add `HeliosDeploymentWorkflowRunSyncService` to synchronize terminal workflow run states from GitHub back into Helios deployments and linked deployments.
- Recover completed workflow runs after failed cancellation attempts instead of surfacing a generic cancellation failure.
- Update orphan deployment recovery to synchronize stuck workflow-backed deployments before only failing deployments that have neither deployment nor workflow run IDs.
- Update deployment timing logic so `REQUESTED`, `WAITING`, `PENDING`, and `QUEUED` stay in the pre-deployment step while timers start only when the deployment actually reaches `PENDING`.
- Add focused server and client tests for synchronization, cancellation recovery, queued-state mapping, deployment timing, and stepper rendering.

#### Prerequisites:
- GitHub Account with Developer access to a Helios repository.
- A configured environment with a deployment workflow.

#### Flow:
1. Log in to Helios as a Developer.
2. Trigger a deployment whose GitHub Actions.
3. Disable ngrok before receiving a completed event for the workflow
4. Cancel a deployment after its GitHub workflow run has already completed and verify Helios synchronizes the terminal status from GitHub instead of keeping the deployment stale.

#### General

#### Server

### DIFF
--- a/server/application-server/src/main/java/de/tum/cit/aet/helios/deployment/HeliosDeploymentWorkflowRunSyncService.java
+++ b/server/application-server/src/main/java/de/tum/cit/aet/helios/deployment/HeliosDeploymentWorkflowRunSyncService.java
@@ -1,0 +1,139 @@
+package de.tum.cit.aet.helios.deployment;
+
+import de.tum.cit.aet.helios.github.GitHubService;
+import de.tum.cit.aet.helios.heliosdeployment.HeliosDeployment;
+import de.tum.cit.aet.helios.heliosdeployment.HeliosDeploymentRepository;
+import de.tum.cit.aet.helios.workflow.WorkflowRun;
+import de.tum.cit.aet.helios.workflow.github.GitHubWorkflowRunStateMapper;
+import java.io.IOException;
+import java.time.OffsetDateTime;
+import java.util.Optional;
+import lombok.RequiredArgsConstructor;
+import lombok.extern.log4j.Log4j2;
+import org.springframework.stereotype.Service;
+import org.springframework.transaction.annotation.Transactional;
+
+@Service
+@RequiredArgsConstructor
+@Log4j2
+public class HeliosDeploymentWorkflowRunSyncService {
+
+  private final GitHubService gitHubService;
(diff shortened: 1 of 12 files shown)


mertilginoglu:
@codex review

server/application-server/src/main/java/de/tum/cit/aet/helios/deployment/HeliosDeploymentWorkflowRunSyncService.java
```diff
@@ -0,0 +1,139 @@
...
+          if (timestampAdvanced) {
+            deployment.setUpdatedAt(workflowRunUpdatedAt);
+          }
+          if (stateChanged || timestampAdvanced) {
+            deploymentRepository.save(deployment);
+          }
+        });
+  }
+
+  private Optional<HeliosDeployment.Status> mapTerminalWorkflowRunState(
```
meryemefe, line 103: I think we can re-use this method: `de.tum.cit.aet.helios.heliosdeployment.HeliosDeployment#mapWorkflowRunStatus` What do you think?

meryemefe (commented):
I only have one comment regarding a reusable method. Apart from that, I tested it. LGTM

PR #1008, opened 2026-05-01
refactor: replace granular log line tone toggles with a consolidated log level filter component

### LINKED ISSUE #985: [Feature]:  Improve workflow log levels and presentation
### Contact Details

_No response_

### Describe the feature

Refine how workflow logs are categorized and displayed to improve clarity and usability.

Requirements:

- [ ] Hide a log level in UI elements if there are no logs in this log level. For example, if there are no warning then don’t show the button in UI.
- [ ] Make the log groups more understandable. In a Workflow Run details, we are displaying jobs. But, in the Workflow Logs, logs groups are based on job steps.

mertilginoglu:
### Motivation

Workflow run logs currently expose several low-level line tone toggles, which makes filtering harder to scan and can leave empty group rows visible when filters hide their children. Reviewers need a simpler way to focus on warnings and errors while keeping grouped log context readable.

Fixes #985 

### Description

- Replaced individual log line tone toggles with a single log level select button.
- Added filter modes for all logs, warnings and errors, and errors only.
- Hide the filter toolbar when the selected log file has no warnings or errors.
- Preserve matching log groups while filtering and skip groups without matching child content.
- Updated workflow run log component and utility tests for the new filtering behavior.

#### Prerequisites:
- A repository in Helios with completed workflow runs
- At least one workflow run log file containing warnings or errors

#### Flow:
1. Log in to Helios as a Developer.
2. Open a repository with completed workflow runs.
3. Navigate to a workflow run logs page.
4. Select a log file that contains warnings or errors.
5. Verify the log level filter is shown above the log output.
6. Switch between `All Logs`, `Warnings & Errors`, and `Errors Only`.
7. Verify each filter shows only the expected log lines while preserving relevant group rows.
8. Select a log file without warnings or errors and verify the filter toolbar is hidden.

### Screenshots

https://github.com/user-attachments/assets/578b5188-0475-4cbe-94b6-808850016164

#### General

#### Client

### DIFF
--- a/client/src/app/pages/workflow-run-logs/workflow-run-logs.component.html
+++ b/client/src/app/pages/workflow-run-logs/workflow-run-logs.component.html
@@ -177,7 +177,7 @@
                     <div class="mt-1 flex min-w-0 flex-wrap items-center gap-x-2 gap-y-1 text-xs text-muted-color">
                       <span class="font-semibold uppercase text-primary-700 dark:text-primary-100">Job</span>
                       <span aria-hidden="true">·</span>
-                      <span>{{ getStepLogCountLabel(groupView.group.files.length) }}</span>
+                      <span>{{ getStepLogCountLabel(getStepFileViewsForGroup(groupView.group.name).length) }}</span>
 
                       @if (groupView.group.jobName && groupView.group.jobName !== groupView.group.name) {
                         <span aria-hidden="true">·</span>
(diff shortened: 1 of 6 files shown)


meryemefe (changes requested):
I tested, LGTM. But I have only one minor comment: 

Here are the steps of a job in my workflow: 

Image

 Here is the screenshot of Helios component:
Image

I think it would be better if we could separate system logs and full logs from step logs. Because, in this screenshot, it says there are 8 steps, but we have actually 7 steps and one of them is skipped. So, we should see 6 steps.

mertilginoglu:
> I tested, LGTM. But I have only one minor comment:
> 
> Here are the steps of a job in my workflow:
> 
> Image
> Here is the screenshot of Helios component: Image
> 
> I think it would be better if we could separate system logs and full logs from step logs. Because, in this screenshot, it says there are 8 steps, but we have actually 7 steps and one of them is skipped. So, we should see 6 steps.

Interesting, in my local steps are properly separated. Does it also happen for other workflows?

meryemefe (approved):

PR #1046, opened 2026-05-18
feat(queue): GitHub Actions queue monitoring, ETA, and SLO alerts

krusche:
## Summary

Adds queue visibility to Helios — per-label-set queue depth, self-hosted runner inventory, build-time percentiles, per-job ETA, stuck-job classification, and SLO alerts — closing the regression Artemis maintainers see vs the previous Bamboo dashboard.

All scheduled jobs + controllers are gated by `helios.queue.enabled` and **default off**. The Flyway migration runs regardless (table creation only); rollback is a forward-fix migration, not a flag flip.

### What's in here
- **DB (`V56__add_workflow_job_and_runner_inventory.sql`)** — `workflow_job`, `runner`, `queue_wait_stat`, `queue_alert_rule`, `queue_alert_event`, with a partial index on `status='queued'`, GIN on labels, and FK to `repository(repository_id)`.
- **Ingestion** — `GitHubWorkflowJobMessageHandler` also persists durable `workflow_job` rows + a Caffeine hot index, **wrapped in try/catch** so a failure cannot poison the existing deployment-timing path. New `GitHubSelfHostedRunnerMessageHandler` for org-level runner events.
- **Reconcilers** (each `@ConditionalOnProperty helios.queue.enabled`; `@EnableScheduling` lives on a `@Profile("!openapi")` config so they never fire during OpenAPI doc generation) — runner inventory (60s), in-progress job filler with `last_reconcile_attempt_at` backoff (30s), stuck-job classifier (60s), hourly p50/p90/p95 rollup (5min), 30-day backfill (admin-triggered, self-throttling to 180 req/min).
- **REST layer** — `EtagCache` + `GitHubRestClient` with `If-None-Match` / 304 reuse and rate-limit metrics.
- **ETA** — label-superset capacity, 3s Caffeine cache, configurable GitHub-hosted concurrency ceiling.
- **Alerts** — `QueueAlertEvaluator` with dedup-via-open-events, a real `HH:mm-HH:mm` quiet window, email channel + template; pluggable `AlertChannel` interface.
- **Controllers** — `WorkflowQueueController` (depth/jobs/stats/alerts CRUD/backfill) and `RunnerController` (`@PreAuthorize("isAuthenticated()")`). Reads are `permitAll` (app-wide GET policy); rule writes require `@EnforceAtLeastWritePermission` **and** that the path `{repoId}` matches the authorized (`X-Repository-Id`) repo.
- **Client** — `ThemeService` extracted from `MainLayoutComponent` so the new `HeliosLineChartComponent` (PrimeNG `<p-chart>` + Chart.js, with `chartjs-adapter-date-fns`) reacts to dark-mode toggles. Routes: `/repo/:id/ci-cd/queue` (overview, runners, stats, alerts) + admin-only top-level `/queue`.
- **Tests** — full server suite **496 tests** green; client suite **221 tests** green.

### Deep review — issues found & fixed

All 7 follow-ups flagged in the first review are now resolved:

1. **Backfill self-invocation** — `start()` dispatches through a proxied `WorkflowJobBackfillExecutor` (so `@Async` runs off-thread), and each page is persisted through a proxied `saveJobPage(@Transactional)` so the transaction is real and wraps only the DB writes (not the GitHub pagination — no connection held across network I/O).
2. **Chart time-axis** — `chartjs-adapter-date-fns` is imported by `HeliosLineChartComponent` and pinned in `package.json`; the stats page no longer throws.
3. **Alert-rule writes 403'd** — the backend derives the WRITE role from `X-Repository-Id`, but `queue.api.ts` (a raw `HttpClient`) never sent it, and `BearerInterceptor` replaced the whole header object (dropping any caller header). Fixed: `BearerInterceptor` now uses `setHeaders`, and the write calls send `X-Repository-Id` matching the path repo. Covered by `queue.api.spec.ts`.
4. **`LabelSets` separator** — was a raw, invisible `0x01` control byte that read as `String.join("", …)` (i.e. as the very collision the comment claimed to prevent). Replaced with a named `""` constant — byte-identical at runtime, so no `labelSetHash` migration, but visible and reformat-safe.
5. **`QueueIndexService` counter drift** — now tracks per-job `JobState`; a redelivered webhook in the same state is a no-op, so the queued counter no longer drifts on retries.
6. **Quiet hours** — `inQuietWindow` parses `HH:mm-HH:mm` as a real window (was a cron fire-moment that only suppressed for one minute).
7. **Stats percentiles** — sample-weighted percentile across buckets (was an unweighted mean of per-bucket p95s).

Additional hardening from a second deep review:

- **Cross-repo write IDOR** — `createRule`/`updateRule`/`deleteRule` now require the path `{repoId}` to equal the authorized (`X-Repository-Id`) repo (403 otherwise); previously a user with write access to repo A could target repo B via the path. New `WorkflowQueueControllerAuthzTest`.
- **Reconcile backoff was a no-op** — `touchReconcileAttempt` / `markMissingOffline` now use `@Modifying(clearAutomatically = true, flushAutomatically = true)`. Inside the `@Transactional reconcile()` loop the saved entities are the same rows the bulk `UPDATE` touches; without flush-before / clear-after, the entity flush at commit clobbered the updated `last_reconcile_attempt_at`, so the same jobs were re-reconciled every 30s (hammering GitHub).
- **Stuck-job classifier** lower-cases both job and runner labels before matching (uppercase `runs-on` labels never matched).
- **Alert email template** uses the plain `${key}` substitution engine (removed FreeMarker `${x!"-"}` / `<#if>` syntax that rendered literally).
- **`/alerts/events` `hoursBack`** clamped to 1h..90d (`findRecent` has no `LIMIT`).

### Remaining decisions (not bugs — a call to make before/within prod enablement)
- **Multi-instance**: the `@Scheduled` reconcilers and alert open-event creation assume a single app instance (no leader election / distributed lock).
- **Quiet-window timezone**: the window is interpreted in the server's local time — decide whose zone should apply.
- **NATS handlers always-ack** (no retry / dead-letter).

### Verification
- `./gradlew :application-server:test` — 496 tests, 0 failures.
- Client unit suite — 221 tests (incl. `queue.api.spec.ts` asserting the write header is sent).
- `./gradlew :application-server:generateOpenApiDocs` + Validate OpenAPI — green.
- `flywayValidate` / `validate-migrations` — `V56` applies cleanly after `V53`–`V55`.
- CI on this PR: server-tests, client-tests, validate-migrations, Validate OpenAPI, lint (server / client / keycloakify), CodeQL — all green (Codacy is non-blocking).

### Still to do
- [ ] Manual staging walkthrough once GitHub App permissions are updated (subscribe `self_hosted_runner`, add org `administration:read`), with `helios.queue.enabled=true`:
  - [ ] webhook job appears in `/ci-cd/queue` within 5s
  - [ ] take a self-hosted runner offline → transitions to OFFLINE within 60s
  - [ ] **create an alert rule** (exercises the write path end-to-end), trigger a low `QUEUE_P95_OVER` → email fires once and auto-clears
  - [ ] POST `/api/queue/admin/backfill` → rate-limit metric stays healthy
- [ ] Decide the three "remaining decisions" above before promoting `helios.queue.enabled=true` to prod.


bassner:
@Claudia-Anthropica review

krusche:
## Deep review (Claude) — updated from staging + 5-agent audit

Merged latest `staging` into the branch (resolved the openapi-gen + `NotificationPreference` conflicts) and fixed a **silent migration collision**: this PR's `V54` clashed with staging's `V54`, and its `chk_notification_type` rewrite **omitted `DEPLOYMENT_APPROVAL_REQUEST`** — applied after staging's V54 it would have dropped that type. Renumbered to **V56** and restored all 7 types. Server 487/488 tests pass (the 1 failure is the pre-existing Docker-provider `WorkflowRunOrphanSweepIntegrationTest`, env-only), client 218 pass.

### ✅ Fixed & pushed (`28fafec`, plus the merge commit)
1. **Scheduling ran under the `openapi` profile** → boot threw `Table WORKFLOW_JOB/QUEUE_WAIT_STAT not found` and the runner reconciler called **live GitHub** during every `generateOpenApiDocs`. Moved `@EnableScheduling` to a `@Profile("!openapi")` config. Verified: 0 scheduled-task errors on boot, queue endpoints still in the spec.
2. **`StuckJobClassifier` matched job labels un-lowercased** against lower-cased runner labels → any uppercase `runs-on` label → false `NO_RUNNER_ONLINE`. Now lower-cases both.
3. **`queue-alert.html` used FreeMarker** (`${x!"-"}`, `<#if>`) but the engine is plain `${key}` regex → measured/threshold rendered blank + literal `<#if>` text. Converted to plain keys.

### 🔴 HIGH — need a decision / careful fix (not auto-fixed; would risk the "make it reliable" goal to do blind)
- **Authorization / cross-repo IDOR.** Queue/runner/alert **GET**s have no repo-scoping — any authenticated user can read any repo's queue depth, job names/branches, percentiles, alert rules (`WorkflowQueueController`), and the whole org's runner inventory (`RunnerController` is `isAuthenticated()` only). `createRule` checks WRITE but not against the **path** `repoId` (`X-Repository-Id` header drives the role) → WRITE on repo A can create a rule on repo B. *The unscoped GET-by-repoId is a pre-existing Helios pattern (`WorkflowController`), so read-scoping is a cross-cutting call; the cross-repo `createRule` write is the most concrete new hole.*
- **Runner reconciler dies/thrashes on ETag 304.** `GitHubRestClient.get()` returns empty on 304; the reconciler treats empty as both "stop paging" and (with `sawAnyResponse`) "mark offline." Once inventory is stable→304, it no-ops forever (a crashed runner with no webhook never gets marked offline); with >100 runners, a 304 on page 2 truncates `seen` and flips healthy page-2 runners OFFLINE. Fix: on 304 return the cached body (`EtagCache.getBody`); don't offline on a truncated page.
- **NATS handlers always `ack` → no retry, silent data loss.** A transient DB error in the `workflow_job` upsert permanently loses that job's final state (the "swallow to avoid redelivery" comment is moot — redelivery never happens). Architectural (affects all handlers): nack/term on transient failure or a DLQ.
- **`QueueIndexService` queue-depth counter desyncs** on 4h state eviction, out-of-order events, and resets to 0 on restart while jobs are still queued. Prefer deriving depth from a DB `COUNT(queued)` (or periodic reconcile) over an in-memory delta counter.
- **Alert open-event dedup is check-then-insert with no conflict handling** → the partial unique index throws under concurrent/multi-instance eval, and the swallow-in-loop can roll back the surrounding `@Transactional`. Use an upsert / catch the constraint.

### 🟡 MEDIUM
- **`/stats` percentiles are a sample-weighted mean of per-bucket percentiles** — not a real window percentile (Javadoc admits "approximation"; the test only proves "beats naive mean"). Heterogeneous buckets inflate/deflate it. Needs raw samples/histograms for accuracy.
- **ETA**: fallback median uses nearest-rank upper-middle (`[10,20]`→20) while the rollup uses `PERCENTILE_CONT` (→15) — divergent; reference lookup **ignores `labelSetHash`** (matrix jobs get an arbitrary label set's p50); model mixes queue models; no staleness bound on the "last 50 completions."
- **`PERCENTILE_CONT` (double) truncated into `INT` columns** — biases stored percentiles low.
- **`@Transactional` reconcilers make GitHub HTTP calls inside the transaction** (`InProgressJobReconciler`, `StuckJobClassifier`) → connection-pool pinning + clobber risk against the webhook path (no `@Version`/lock; a stale reconciler save can revert `completed`→`queued`).
- **`WorkflowJobBackfillService.ingestRunJobs` `@Transactional` is dead** (self-invoked private) → backfill isn't atomic per run; unguarded `OffsetDateTime.parse` can abort the loop.
- **`markMissingOffline` flips ALL runners offline on a single empty/blip GitHub response** — needs an N-consecutive-miss guard.
- **`V56` adds the 3 queue notification types to the CHECK but no preference backfill** — *product call*: default-ON for all users would mass-email on every breach; opt-in means no emails until a user enables it (and an already-"open" event then won't re-fire). Decide the intended default + fix the open-event-suppresses-refire interaction.
- **Quiet-hours uses unanchored `LocalTime.now()`** (host TZ/DST shifts the window); **`measureQueueP95` 5-min window vs hourly buckets** (alert lingers after the queue drains / false-resolves).
- **`LabelSets.hash` joins with an empty separator** → `["x","y"]` and `["xy"]` collide (latent; no test covers it — the "adjacency test" one agent cited doesn't exist).
- **`WorkflowYamlCache` negative-caches transient failures for 1h**; **`GitHubRestClient`** ignores `Link` pagination, can't distinguish 403/429 from 304, no `Retry-After`, `rateLimitRemaining()` is a `-1` stub.
- **`/stats` & `/events` unbounded** (no clamp on `hoursBack`, no `LIMIT`); `runner-list` polls every 5s org-wide; `jobs` runs ETA per row (N+1, 3s-cached).
- **Self-hosted runner handler** can create a phantom row from a late `deleted`/`offline` event and overwrite fresh data with stale (no event-timestamp guard).

### ✔️ Verified correct (so you don't re-check)
V56 columns/types/lengths/nullability/CHECK all match the entities; no SQL injection; `AlertRuleDto` validation is solid; no `[innerHTML]` XSS in the client; `EtagCache` is thread-safe + bounded (Caffeine); `WorkflowJobPersistenceService` upsert is null-safe + clamps negatives; `QueueIndexService` redelivery (same-state) is idempotent.

I fixed the three clear, self-contained bugs that affect boot/CI and obviously-wrong behavior. The HIGH items above are either pre-existing-pattern (authz) or need an architectural/product decision (NATS retry, counter source-of-truth, alert backfill default) — happy to implement any of them once you choose the direction; the 304-reconciler and dead-`@Transactional` ones I can fix straight away if you want them in this PR.

server/application-server/src/main/java/de/tum/cit/aet/helios/workflow/queue/alert/QueueAlertEvaluator.java
```diff
@@ -0,0 +1,190 @@
...
+  private Integer measure(QueueAlertRule rule) {
+    return switch (rule.getKind()) {
+      case QUEUE_P95_OVER -> measureQueueP95(rule);
+      case RUNNER_OFFLINE_OVER -> measureRunnersOffline(rule);
+      case STUCK_JOBS_OVER -> measureStuckJobs(rule);
+    };
+  }
+
+  private Integer measureQueueP95(QueueAlertRule rule) {
+    OffsetDateTime since = OffsetDateTime.now().minusMinutes(rule.getWindowMinutes());
```
Claudia-Anthropica, line 119: @krusche [high] This window is in minutes, but the data source is hourly rollup rows keyed by `bucket_start`. With the default 5-minute window, `since` will almost always be after the latest closed-hour bucket, so `QUEUE_P95_OVER` rules never measure anything and never fire. Either evaluate this alert from raw `workflow_job` rows for short windows, or align the query to buckets that overlap the requested window and enforce a minimum rollup-backed window.

In `server/application-server/src/main/java/de/tum/cit/aet/helios/workflow/queue/alert/QueueAlertEvaluator.java`, `measureQueueP95` compares a minute-level alert window against hourly `queue_wait_stat.bucket_start` values, so default 5-minute rules usually see no data. Fix this by evaluating p95 from raw completed workflow jobs for the requested window or by querying hourly buckets that overlap the window and rejecting or normalizing unsupported short windows.

</details>

server/application-server/src/main/java/de/tum/cit/aet/helios/workflow/queue/reconcile/RunnerInventoryReconciler.java
```diff
@@ -0,0 +1,110 @@
...
+        }
+        runner.setLastSeenAt(now);
+        runnerRepository.save(runner);
+      }
+      if (runners.size() < perPage) {
+        break;
+      }
+      page++;
+    }
+    if (sawAnyResponse) {
```
Claudia-Anthropica, line 101: @krusche [high] `sawAnyResponse` is true after page 1, so a transient failure or 403 on page 2 still falls through and marks every runner not seen on the earlier pages as offline. In orgs with more than 100 runners this can create false offline inventory and false alerts. Only call `markMissingOffline` after the pagination completed successfully; on any page fetch failure, keep the previous inventory.

In `server/application-server/src/main/java/de/tum/cit/aet/helios/workflow/queue/reconcile/RunnerInventoryReconciler.java`, a partial paginated runner inventory fetch can still mark unseen later-page runners offline. Fix this by tracking whether all pages were fetched successfully and calling `markMissingOffline` only after a complete inventory response; skip the offline sweep on any page fetch failure.

</details>

server/application-server/src/main/java/de/tum/cit/aet/helios/workflow/queue/reconcile/WorkflowJobBackfillService.java
```diff
@@ -0,0 +1,239 @@
...
+    } finally {
+      running.set(false);
+    }
+  }
+
+  public void ingestRunJobs(String fullName, Long runId, Long repositoryId,
+      String workflowName, String headBranch, String headSha) {
+    int page = 1;
+    while (!aborted.get()) {
+      String path =
```
Claudia-Anthropica, line 130: @krusche [medium] The 180 req/min throttle only wraps the `/actions/runs` page request; each run then calls `/jobs` immediately inside `ingestRunJobs`. A 100-run page can therefore issue 100+ extra REST calls back-to-back, hit GitHub rate limits, and leave the backfill partial. Apply the same throttle/rate limiter to every REST call, including job pages.

In `server/application-server/src/main/java/de/tum/cit/aet/helios/workflow/queue/reconcile/WorkflowJobBackfillService.java`, the backfill throttle is only applied to workflow-run page requests and not to per-run job page requests. Fix this by applying one shared throttle or rate limiter before every GitHub REST call made by the backfill, including calls inside `ingestRunJobs`.

</details>

server/application-server/src/main/java/de/tum/cit/aet/helios/workflow/queue/reconcile/QueueWaitStatRollup.java
```diff
@@ -0,0 +1,102 @@
...
+          bucketEnd);
+    }
+  }
+
+  /** Rolls up every hour bucket from {@code from} (inclusive) to {@code until} (exclusive). */
+  public void rollupRange(OffsetDateTime from, OffsetDateTime until) {
+    OffsetDateTime cursor = from.truncatedTo(ChronoUnit.HOURS);
+    OffsetDateTime end = until.truncatedTo(ChronoUnit.HOURS);
+    while (cursor.isBefore(end)) {
+      rollupBucket(cursor, cursor.plusHours(1));
```
Claudia-Anthropica, line 98: @krusche [medium] `rollupRange()` is used by the async backfill, but it calls the `@Transactional` method on the same bean. That bypasses the Spring proxy, so the native `executeUpdate()` can run without a transaction and the backfill's historical stats materialization fails. Make `rollupRange` transactional itself or invoke `rollupBucket` through a proxied bean/`TransactionTemplate`.

In `server/application-server/src/main/java/de/tum/cit/aet/helios/workflow/queue/reconcile/QueueWaitStatRollup.java`, `rollupRange()` self-invokes the `@Transactional rollupBucket()` method, so backfill-triggered native updates can run outside a transaction. Fix this by giving `rollupRange()` its own transaction boundary or by executing each bucket through a proxied bean or `TransactionTemplate`.

</details>

client/src/app/pages/queue/queue-alerts/queue-alerts.component.ts
```diff
@@ -0,0 +1,196 @@
...
+          </ng-template>
+          <ng-template pTemplate="body" let-rule>
+            <tr>
+              <td>{{ rule.kind }}</td>
+              <td>{{ rule.thresholdSeconds }}s</td>
+              <td>{{ rule.windowMinutes }}m</td>
+              <td>
+                <p-tag [value]="rule.enabled ? 'on' : 'off'" [severity]="rule.enabled ? 'success' : 'secondary'" />
+              </td>
+              <td>{{ rule.quietHoursCron ?? '—' }}</td>
```
Claudia-Anthropica, line 74: @krusche [low] The DTO now exposes `quietWindow`, but the table still reads the old `quietHoursCron` property. Existing rules with a quiet window will always render as `—`, which makes it look like the setting was lost. Bind this cell to `rule.quietWindow`.

In `client/src/app/pages/queue/queue-alerts/queue-alerts.component.ts`, the alert rules table reads `rule.quietHoursCron` even though the DTO field is `quietWindow`. Fix the template to display `rule.quietWindow ?? '—'`.

</details>

Claudia-Anthropica (changes requested):
@krusche The implementation has a lot of the earlier hardening in place, but I found a few correctness holes that would make queue alerts/backfill lie or silently fail. The biggest ones are the p95 alert window querying hourly buckets incorrectly and runner inventory marking later pages offline after a partial GitHub response; see inline. Codacy is also currently red, so this needs another pass before merge.

PR #1049, opened 2026-05-25
fix(nats): bound JetStream and never reject GitHub webhooks

krusche:
## Background

Twice in recent months, the JetStream `nats-data` volume filled the host disk and `nats-server` refused every webhook publish until it was manually rescued. Root cause: the streams had no size cap, the server had no JetStream-wide ceiling, and the consumer could pile in-flight backlog onto the stream without bound. There was no layer that said *stop*.

This PR adds those layers, while preserving the property that we **never reject a fresh GitHub webhook** — losing inbound events to GitHub's retry exhaustion is worse than losing buffered backlog.

## Final design (this PR vs `staging`)

Four layers of defence, top to bottom:

| Layer | On `staging` today | After this PR |
|---|---|---|
| **Stream policy** | No `max_bytes`, `max_age=28d`, default discard | `max_bytes=2 GiB` (github) / `64 MiB` (notification), `max_age=7d`, explicit `retention=Limits` + `DiscardPolicy.OLD` |
| **Consumer cap** | No `maxAckPending` set (NATS default: 1000) | `maxAckPending=500`, env-tunable via `NATS_CONSUMER_MAX_ACK_PENDING` |
| **JetStream server** | `jetstream { store_dir: /data }` — unbounded | + `max_mem: 1G`, `max_file: 8G` cluster-wide ceiling |
| **Container logging** | Only `nats-server` had `json-file` rotation (10m × 3) | `x-logging` anchor applied to every service (20m × 5) |

The stream bootstrap continues to apply the desired config on every deploy (`add_stream`, falling through to `update_stream` if the stream already exists), so the new `DiscardPolicy.OLD` takes effect automatically on first deploy — no manual operator step.

### Why `DiscardPolicy.OLD` for inbound webhooks

A full stream has to do *something* when the next publish arrives. The two options:

- **`DiscardPolicy.NEW`** rejects the publish. The webhook-listener returns 5xx to GitHub, GitHub retries ~3 times over ~30s, then drops the delivery permanently. We lose the **freshest, most valuable** events first.
- **`DiscardPolicy.OLD`** drops the oldest already-buffered events to make room. The publish always succeeds; GitHub never sees a failure. Under catastrophic consumer lag we lose the **oldest unprocessed** events, which the consumer was going to catch up to eventually anyway.

For webhook ingestion the second is unambiguously correct.

### Why a JetStream-wide `max_file`

The per-stream `max_bytes` is one layer; on its own it would still be the only ceiling in the stack. Anyone adding a new stream tomorrow without `max_bytes`, or any drift on the existing ones, refills the disk. `max_file: 8G` on the JetStream block is a single server-wide hard limit that catches every stream — including ones we haven't written yet. Defence in depth, not a substitute.

### Why `maxAckPending=500`

Consumer fall-behind — not raw inflow — is what causes a stream to approach its cap. The previous default (1000) let a stalled consumer hold a thousand unacked messages in flight, all counted against the stream. Halving that bounds the in-flight pressure and makes consumer lag visible long before the stream ceiling is reached.

## Defence in depth

```
GitHub webhook
      │
      ▼
webhook-listener  ── DiscardPolicy.OLD ──►  publish always succeeds
      │
      ▼
JetStream stream  ── max_bytes (2 GiB) ──►  per-stream cap, sheds oldest
      │
      ▼
JetStream server  ── max_file (8 GiB)  ──►  cluster-wide hard backstop
      │
      ▼
Durable consumer  ── maxAckPending (500) ─►  bounded in-flight pressure
```

If any one layer is misconfigured, the next layer prevents the disk-full incident.

## Test plan

- [ ] `webhook-listener` boots cleanly against an empty NATS server and creates both streams with the expected config.
- [ ] `webhook-listener` boots against an existing drifted stream and applies the new config without manual intervention.
- [ ] `nats stream info github` reports `Discard Policy: Old`, `Maximum Bytes: 2.0 GiB`, `Maximum Age: 7d`.
- [ ] `nats stream info notification` reports the same with 64 MiB.
- [ ] `nats server info` reports JetStream `max_file: 8 GiB`, `max_mem: 1 GiB`.
- [ ] `nats consumer info github <consumer>` reports `Max Ack Pending: 500`.
- [ ] End-to-end: a GitHub webhook reaches `application-server` and triggers the expected downstream effect.

## References

- [NATS JetStream model deep dive](https://docs.nats.io/using-nats/developer/develop_jetstream/model_deep_dive)
- [nats-server discussion #4499 — retention for fan-out](https://github.com/nats-io/nats-server/discussions/4499)
- [nats-server issue #2795 — DiscardNew recovery](https://github.com/nats-io/nats-server/issues/2795)


FelixTJDietrich:
I updated the PR with some of the outage fixes from Hephaestus, which also includes the `maxAckPending` and `DiscardPolicy.OLD` which is the way to go to not let GitHub drop webhooks. 

I also came across the "interest" retention policy which might be of interest in the future for retention so the data only stays in the NATS until it has been successfully processed. But I also quite like the replayability of not immediately deleting it. 

From the code it looks fine to me but I did not manually test it.

PR #1098, opened 2026-05-28
feat(approval): reviewer-aware auto-approval + in-app approve/decline for protected envs

krusche:
### Motivation

Today, when Helios dispatches a deployment to a GitHub environment with a required-reviewer protection rule, three things go wrong:

1. The webhook handler blindly impersonates the *deployment creator* and tries to approve, regardless of whether they're actually a required reviewer. When they aren't, GitHub returns 403 and the error is silently swallowed — the deployment sits in WAITING with no signal to the user.
2. The only way to actually resolve such a pending deployment is to leave Helios, open GitHub's UI, find the workflow run, and click Review pending deployments. Reviewers forget; valuable time is lost.
3. The Keycloak SSO session is hard-capped at 8h, so users effectively need to re-auth roughly daily — independent of the approval problem but symptomatic of the same "leaves Helios to do basic things" pain.

This PR is **Phase 1 + Phase 2** of a focused approval flow that keeps everything inside Helios. Phase 3 (email one-click links) is intentionally out of scope and can land later.

#### Reviewer-aware auto-approval (Phase 1)

When GitHub fires `deployment_status: waiting` for a Helios-originated deployment, `ApprovalService.reviewDeployment` now branches on the env's protection rule:

| Situation | Outcome | `HeliosDeployment.autoApprovalDecision` |
|---|---|---|
| No required-reviewer rule | Do nothing — GitHub handles any wait-timer / branch policy itself | `NOT_APPLICABLE` |
| Team-type reviewers present | Legacy "impersonate deployer, try anyway" (Helios doesn't yet expand teams to members; punting silently would regress working setups) | `TEAM_REVIEWER_FALLBACK` |
| `preventSelfReview` is on, or deployer not in User-type reviewer list | Defer; surface in the in-app pending-approvals list for someone else to resolve | `DEFERRED_TO_REVIEWERS` |
| Deployer **is** a User-type required reviewer | Auto-approve as them via Keycloak impersonation, with a source-tagged comment so the GitHub audit trail says `Auto-approved by Helios (deployer is required reviewer)` | `AUTO_APPROVED` |

A bounded retry (up to 3 attempts, 100 ms + 300 ms backoff = 400 ms worst case) handles the race where the webhook arrives before the deploy-side transaction commits. The service is idempotent against webhook redelivery: once `autoApprovalDecision` is non-null, a second invocation is a no-op.

#### In-app approve / decline (Phase 2)

Two endpoints, gated on "current user is in this env's User-type required reviewer list":

- `POST /api/deployments/{id}/approve` — impersonates the caller, approves on GitHub, marks any sibling pending requests `CONSUMED_BY_OTHER` in the same transaction.
- `POST /api/deployments/{id}/decline` body `{comment?}` — same but `state: rejected` (the workflow run is killed for everyone; the UI surfaces that warning explicitly).
- `GET /api/deployments/pending-approvals` — caller's outstanding requests, joined with deployment + env + repo summary.

Concurrency: `SELECT FOR UPDATE` on the `HeliosDeployment` row serialises concurrent clicks; only one reviewer's GitHub call ever lands. `FAILED_AT_GITHUB` distinguishes "clicked Approve" from "GitHub accepted Approve", letting the UI offer a clean retry. The transaction is annotated `noRollbackFor = ResponseStatusException.class` so the FAILED_AT_GITHUB audit row survives the 502 throw — without this, Spring's default rollback would silently discard the row.

#### `DeploymentApprovalRequest` entity (Phase 1 → 3)

One row per (deployment, reviewer) tracking the lifecycle (`PENDING → APPROVED | DECLINED | EXPIRED | CONSUMED_BY_OTHER | FAILED_AT_GITHUB`), tagged with `via` in `{AUTO, IN_APP, EMAIL_LINK}`. Auto rows are created in terminal state. Email-link Phase 3 will populate `tokenHash` (SHA-256 of the link token, plaintext only in the URL). Flyway V53 adds the table plus `auto_approval_decision` / `auto_approval_at` on `helios_deployment` for fast UI lookups + at-a-glance audit. `ApprovalDecisionWriter` atomically commits the audit row + deployment stamp together so a crash between the two writes cannot leave a half-state.

#### UI

New cross-repo `/pending-approvals` page (Approve / Decline per row, decline opens a dialog with an optional reason and the "this kills the workflow for everyone" warning). Badge on the profile button + a Pending approvals entry in the profile popover, both driven by the same 30 s-polled query — visible across every page since the queue is cross-repo.

#### Realm session lifespans

`ssoSessionMaxLifespan` 8 h → 30 d, `ssoSessionIdleTimeout` 8 h → 7 d on `helios-example-realm.json`. **Prod Keycloak must mirror these** for the user-visible effect; the example file is only the dev/staging seed.

#### GitHubService surface

`approveDeploymentOnBehalfOfUser` gains a `comment` parameter so the source of every approval (auto / in-app / email-link) is captured in the GitHub-side audit comment. A new `rejectDeploymentOnBehalfOfUser(...)` implements the decline path. The legacy 4-arg overload still exists. Token-exchange failure now throws `IOException` instead of silently returning, so callers can react.

### Intentionally NOT in this PR

- **Email one-click approve / decline links** (Phase 3). The plan lays out a token table (`token_hash` is already on `DeploymentApprovalRequest`, ready for use), but no `Notification` payload or template yet.
- **Team-type reviewer expansion** (Phase 4). Today's "impersonate deployer, try anyway" fallback runs when team-type reviewers are present; this preserves existing setups rather than silently no-opping. Phase 4 will resolve team membership via `GET /orgs/{org}/teams/{slug}/members` (cached).
- **Pre-deploy confirmation modal**. Mentioned in the brief; deferred. Phase 2's pending-approvals queue plus the source-tagged audit comments already cover the core requirement ("don't make me leave Helios").
- **`deployment_review` webhook sync**. Approvals that happen outside Helios in the GitHub UI won't sync back to the pending list until Phase 4. The list is correct for Helios-driven deployments, which is the bulk of usage.
- **Admin bypass**. GitHub allows repo admins to bypass reviewer rules in its UI; Helios does not replicate this in code.

#### Prerequisites

- Helios dev stack running locally (server on `:8080`, client on `:4200`, Postgres, Keycloak, NATS, GitHub-app webhook reachable).
- A connected GitHub repository in Helios with **at least one environment that has a `required-reviewer` protection rule** of the **User** type (Settings → Environments → … → Deployment protection rules → Required reviewers).
- Add your GitHub user as a required reviewer of that environment. For the cross-user tests, also add a second user (or use a colleague's account in a second browser profile).
- Make sure your Helios login has gone through the GitHub IDP at least once so Keycloak holds a GitHub token for you (Keycloak Admin Console → *realm* → Users → *you* → Identity Provider Links → `github` should show a stored token).
- Have a trivial deployment workflow on the repo's default branch — anything that targets the protected environment (a `sleep 5` job is fine) — so you don't need to wait for a long build between attempts.

Quick DB peek commands used below (adjust connection params for your dev stack):

```bash
# Latest deployment-approval audit rows
psql -h localhost -U helios -d helios -c \
  "select id, helios_deployment_id, state, via, reviewer_login, responded_at, failure_reason
   from deployment_approval_request order by id desc limit 5;"

# Latest helios_deployment auto-approval decisions
psql -h localhost -U helios -d helios -c \
  "select id, status, auto_approval_decision, auto_approval_at
   from helios_deployment order by id desc limit 5;"
```

#### Server unit tests

```bash
cd server && ./gradlew :application-server:test \
  --tests "*ApprovalServiceTest" \
  --tests "*DeploymentReviewActionServiceTest" \
  --tests "*GitHubServiceTest"
```

These exercise: deployer-is-reviewer auto-approve, defer when `preventSelfReview`, defer when deployer not reviewer, no-rule no-op, team-fallback path, GitHub failure → `FAILED_AT_GITHUB`, persist/webhook race retry, missing `workflowRunId` skip, **idempotency on replayed webhook**, in-app happy-approve / happy-decline / not-a-reviewer-403 / `preventSelfReview`-403 / already-resolved-409 / no-runId-409 / no-rule-409 / GitHub-fail-502 / sibling-`CONSUMED_BY_OTHER`. (208 server tests total; all green.)

#### End-to-end test matrix

The matrix below is what a reviewer should walk through to confirm the feature actually does what the PR promises. Each test takes ~1 min once the prerequisites are in place.

##### Test 1 — Auto-approval when the deployer is a required reviewer (the headline)

**Setup:** the protected env's required reviewers include your account; `preventSelfReview` is **off**.

1. In Helios, open *Repo → Environments* and click **Deploy** on the protected environment.
2. Wait for the workflow run to reach the protected job (a few seconds with a trivial workflow).

**Expected:**
- The GitHub workflow run **resumes automatically within ~1 s of hitting the `waiting` state** — no clicks needed.
- On GitHub: open the workflow run → "Review pending deployments" log shows comment `Auto-approved by Helios (deployer is required reviewer)`.
- DB: `helios_deployment.auto_approval_decision = AUTO_APPROVED`, `auto_approval_at` set; one new `deployment_approval_request` row with `state=APPROVED, via=AUTO, reviewer_login=<you>`.
- Server log: `Auto-approved deployment {id} on behalf of @<you> (AUTO_APPROVED)`.

##### Test 2 — Deferral + in-app Approve (the core Phase 2 path)

**Setup:** the protected env's required reviewer is **someone else** (not you). Easiest: temporarily replace yourself as required reviewer with a colleague's account, or use a second test account in a second browser profile.

1. As **you** (not a reviewer), click **Deploy**.
2. Wait until the workflow reaches the protected job.
3. Switch to the **reviewer's** browser. Click the profile avatar (top-right).

**Expected at the reviewer's side:**
- A warning badge appears on the profile button (within ~30 s — that's the poll interval).
- The dropdown shows a **Pending approvals** entry with the same badge count.
- Clicking it lands on `/pending-approvals` and the deployment is in the table with env, repo (clickable), the deployer's login, branch, commit SHA, and a link to the GitHub run.

4. As the reviewer, click **Approve** on that row → confirm in the dialog.

**Expected:**
- The mutation returns 200; a green toast says "Approved".
- On GitHub the run resumes; the "Review pending deployments" log shows `Approved by @<reviewer> via Helios (in-app)`.
- DB: the audit row `state=APPROVED, via=IN_APP, responded_at` set; `helios_deployment.auto_approval_decision=DEFERRED_TO_REVIEWERS` (the original deferral decision is preserved as the audit anchor).
- The badge on the reviewer's profile clears within ~30 s.

##### Test 3 — In-app Decline rejects the workflow for everyone

**Setup:** same as Test 2 — a pending approval the reviewer can act on.

1. As the reviewer, click **Decline** on a row.
2. The dialog should warn explicitly: "Declining will reject the workflow run on GitHub — the deployment will not happen and the run is marked as failed."
3. Optionally type a reason ("wrong branch"). Click **Decline deployment**.

**Expected:**
- The dialog closes; a warn-severity toast says "Declined".
- The GitHub workflow run goes to **failure / rejected** state (not just paused).
- GitHub's "Review pending deployments" log shows `Declined by @<reviewer> via Helios (in-app): wrong branch` (or without `: …` if no reason).
- DB: `state=DECLINED, via=IN_APP, responded_at` set.

##### Test 4 — `preventSelfReview` honored

**Setup:** the env's required reviewers include your account; `preventSelfReview` is **on**.

1. As yourself, click **Deploy**.
2. Wait for the `waiting` event.

**Expected:**
- GitHub run **does not** resume — no auto-approval.
- DB: `auto_approval_decision = DEFERRED_TO_REVIEWERS`.
- `/pending-approvals` for **you** does **not** show this row (you can't self-review).
- If a second required reviewer exists, they can resolve it through the same Test-2 / Test-3 paths.

##### Test 5 — Not-a-required-reviewer can't approve

As a logged-in user who is **not** in the env's required-reviewer list, attempt the API directly:

```bash
curl -i -X POST http://localhost:8080/api/deployments/{deployment-id}/approve \
  -H "Authorization: Bearer $KEYCLOAK_BEARER"
```

**Expected:** `HTTP 403 Forbidden`, body contains `You are not a required reviewer for this environment.` The UI doesn't expose the button to non-reviewers, so this guards direct API misuse.

##### Test 6 — Already-resolved → 409 Conflict

After Test 2 succeeds, hit the same deployment id again:

```bash
curl -i -X POST http://localhost:8080/api/deployments/{same-id}/approve \
  -H "Authorization: Bearer $KEYCLOAK_BEARER"
```

**Expected:** `HTTP 409 Conflict`, body contains `This deployment has already been resolved.` Also confirms the first-responder race protection: if two reviewers had clicked at the same time, the slower one would see this exact response.

##### Test 7 — Idempotency on webhook redelivery (the subtle one)

Setup: a deployment from Test 1 (already `AUTO_APPROVED`).

1. GitHub → repo Settings → Webhooks → (Helios webhook) → Recent Deliveries → find the `deployment_status` event with `state=waiting` for your run → **Redeliver**.

**Expected:**
- The Helios server log shows `Skipping reviewDeployment for {id}; already decided as AUTO_APPROVED.`
- No new `deployment_approval_request` row is created (still exactly one APPROVED row).
- `helios_deployment.auto_approval_decision` is **not** overwritten — it stays `AUTO_APPROVED`.
- No new GitHub API call (look for absence of an `approveDeploymentOnBehalfOfUser` log line).

##### Test 8 — Env without required reviewers is left alone

**Setup:** deploy to an env that has **no** required-reviewer rule (a wait-timer-only env, or no protection rules at all).

1. Click Deploy.

**Expected:**
- GitHub handles the deployment per its own rules (wait-timer expires, or no gate).
- Helios stamps `auto_approval_decision = NOT_APPLICABLE`.
- No `deployment_approval_request` rows are created.
- Server log: `No required-reviewer rule for environment {name}; leaving deployment {id} to GitHub.`

#### Visual / UX smoke

- The `/pending-approvals` page handles **empty list** ("Nothing waiting on you right now.") and loading skeleton correctly.
- The Approve / Decline buttons disable while a mutation is in flight (no double-submits possible by spam-clicking).
- The decline dialog blocks until the user types a reason **or** clicks "Decline deployment" with no reason — both should work; the GitHub-side comment changes accordingly.
- The badge on the profile button updates within ~30 s of the underlying list changing (refetch interval).

### Operator notes

- **Mirror the realm session lifespans on prod Keycloak**: Admin Console → Realm Settings → Sessions → `SSO Session Idle: 7 d`, `SSO Session Max: 30 d`. The realm-export json in this PR is only the dev/staging seed.
- Confirm the GitHub IDP on the prod realm still has `Store Tokens` + `Stored Tokens Readable` enabled (these are baked into the example realm; verify on prod).
- `helios.clientBaseUrl` per profile is used today by the email pipeline; no change here, but Phase 3 will rely on it.


krusche:
@Claudia-Anthropica review

Claudia-Anthropica:
@krusche on it! Taking a look now :eyes:

server/application-server/src/main/java/de/tum/cit/aet/helios/deployment/approval/ApprovalService.java
```diff
@@ -2,61 +2,330 @@
...
+    if (heliosDeployment == null) {
+      // The deployment_status webhook arrived for something Helios doesn't know about — most
+      // likely a deployment that wasn't dispatched from Helios. Leave it for GitHub's normal flow.
       log.info("Skipping approval for non-Helios deployment with ID {}", deploymentSource.getId());
       return;
     }
 
-    HeliosDeployment heliosDeployment = heliosDeploymentOpt.get();
+    // Idempotency: GitHub redelivers webhooks; we treat the first decision as authoritative.
+    AutoApprovalDecision priorDecision = heliosDeployment.getAutoApprovalDecision();
```
Claudia-Anthropica, line 78: @krusche [high] This idempotency check only handles redeliveries after the first invocation has committed. Two `deployment_status: waiting` deliveries for the same deployment can both read `null` here, both call GitHub, and then the later writer can overwrite an `AUTO_APPROVED` decision with a failure/fallback row if GitHub rejects the second call. Serialize the decision claim per deployment, or use a conditional update, before making the GitHub review call.

In `server/application-server/src/main/java/de/tum/cit/aet/helios/deployment/approval/ApprovalService.java`, concurrent waiting webhooks can both pass the `autoApprovalDecision == null` check and double-act on GitHub. Fix this by serializing approval-decision processing per `HeliosDeployment` with a row lock or compare-and-set update before the GitHub call, and add coverage for concurrent/redelivered waiting events.

</details>

server/application-server/src/main/java/de/tum/cit/aet/helios/deployment/approval/ApprovalService.java
```diff
@@ -2,61 +2,330 @@
...
+              : (creatorLogin == null
+                  ? "no creator login recorded"
+                  : "creator is not in required-reviewer list");
+      log.info(
+          "Deferring deployment {} ({}): {} required reviewers will be notified [{}].",
+          heliosDeployment.getId(),
+          reason,
+          reviewers.userLogins().size(),
+          String.join(", ", reviewers.userLogins()));
+      decisionWriter.recordDecisionOnly(
```
Claudia-Anthropica, line 162: @krusche [medium] This stamps the deployment as already decided before the pending reviewer rows are saved. If the process dies, or any DB write in `notifyReviewers` fails, a later webhook redelivery will stop at the idempotency guard and nobody gets a pending approval row for a still-waiting deployment. Persist the deferred decision and the reviewer rows atomically.

In `server/application-server/src/main/java/de/tum/cit/aet/helios/deployment/approval/ApprovalService.java`, the `DEFERRED_TO_REVIEWERS` decision is saved before the pending approval rows are created. Fix this by saving the deferred decision and all reviewer `DeploymentApprovalRequest` rows in one transaction, so redelivery cannot skip a deployment with no visible approval requests.

</details>

server/application-server/src/main/java/de/tum/cit/aet/helios/deployment/approval/DeploymentApprovalRequestRepository.java
```diff
@@ -0,0 +1,57 @@
...
+   * inlined form needs the non-portable nested-class {@code $} separator and is brittle to
+   * Hibernate parser changes.
+   */
+  @Query(
+      "select r from DeploymentApprovalRequest r "
+          + "join fetch r.heliosDeployment d "
+          + "left join fetch d.creator "
+          + "join fetch d.environment e "
+          + "join fetch e.repository "
+          + "where r.reviewer.id = :reviewerId and r.state = :state "
```
Claudia-Anthropica, line 51: @krusche [high] `FAILED_AT_GITHUB` rows never come back to the retry surface. `DeploymentReviewActionService` writes that state after a transient GitHub failure, but this query only returns `PENDING`, so the reviewer sees an empty queue while GitHub is still waiting. Include failed rows in the pending-approvals response and let the UI retry, or keep the row pending with failure metadata.

In `server/application-server/src/main/java/de/tum/cit/aet/helios/deployment/approval/DeploymentApprovalRequestRepository.java`, the pending approvals query excludes `FAILED_AT_GITHUB` rows even though those rows need a retry path. Fix this by returning retryable failed rows to `/api/deployments/pending-approvals` and updating the DTO/UI/service logic so reviewers can retry after a GitHub failure.

</details>

Claudia-Anthropica (changes requested):
@krusche The feature direction looks solid, but the approval flow still has a couple of correctness gaps around retries and webhook redelivery. CI is also still failing because the generated OpenAPI spec differs from `openapi.yaml` for the new notification enum.

server/application-server/src/main/java/de/tum/cit/aet/helios/deployment/approval/ApprovalService.java
Claudia-Anthropica: @krusche Still open: `reviewDeployment` still reads `autoApprovalDecision` without taking a lock or doing a conditional claim before the GitHub call, so two waiting webhooks can both pass this guard and double-act on GitHub. The decision needs to be serialized before any external approval/rejection attempt.

server/application-server/src/main/java/de/tum/cit/aet/helios/deployment/approval/ApprovalService.java
Claudia-Anthropica: @krusche Still open: the `DEFERRED_TO_REVIEWERS` stamp is still saved by `recordDecisionOnly(...)` before `notifyReviewers(...)` creates the pending rows. If that second step fails or the process dies, redelivery skips the deployment because the decision is already set, but reviewers still have no visible request.

server/application-server/src/main/java/de/tum/cit/aet/helios/deployment/approval/DeploymentApprovalRequestRepository.java
Claudia-Anthropica: @krusche Still open: `/api/deployments/pending-approvals` still calls `findByReviewerAndState(..., PENDING)`, so a row moved to `FAILED_AT_GITHUB` disappears from the retry queue. The retry surface needs to include those failed rows or keep the row pending with failure metadata.

Claudia-Anthropica (changes requested):
@krusche The previous blockers are still open in the current code, so I am keeping this at changes requested. The approval decision still is not serialized before the GitHub call, deferred reviewer rows are still not persisted atomically with the decision stamp, and failed GitHub attempts still disappear from the pending-approvals retry queue.

server/application-server/src/main/java/de/tum/cit/aet/helios/deployment/approval/ApprovalService.java
krusche: Fixed in dfbe636. `ApprovalDecisionWriter` now re-loads the deployment under `SELECT ... FOR UPDATE` and re-checks that `autoApprovalDecision` is still null before writing, so a concurrent in-app approve/decline (which holds the row lock for its whole transaction) can no longer overwrite a recorded decision or write a duplicate audit row — the first recorded decision wins and a losing writer becomes a no-op. (webhook-vs-webhook is already serialized by the single-threaded JetStream consumer; the real window was webhook-vs-in-app, now closed.)

server/application-server/src/main/java/de/tum/cit/aet/helios/deployment/approval/ApprovalService.java
krusche: Fixed in dfbe636. The deferral branch now persists the per-reviewer PENDING rows (and publishes the emails) BEFORE stamping `DEFERRED_TO_REVIEWERS`. Since the stamp is the idempotency gate, stamping last means a failure in `notifyReviewers` leaves `autoApprovalDecision` null so a redelivered webhook re-runs the whole deferral instead of short-circuiting and stranding the deployment.

server/application-server/src/main/java/de/tum/cit/aet/helios/deployment/approval/DeploymentApprovalRequestRepository.java
krusche: Fixed in dfbe636. `GET /api/deployments/pending-approvals` now returns both `PENDING` and `FAILED_AT_GITHUB` rows (new `findByReviewerAndStateIn` join-fetch query), so a row that failed at GitHub reappears in the reviewer's queue (and the profile badge, which counts the list) and clicking Approve/Decline reuses the same row to retry.

krusche:
## Deep review (Claude) — 1 fix pushed, remaining findings for your call

I ran a focused deep review (server approval logic, GitHub integration + migrations, client UI) plus re-verified the three already-resolved threads. Summary below.

### ✅ Fixed in this push (`0e5b594`)
- **[HIGH] Stored XSS in the approve confirmation dialog** — `pending-approvals.component.ts` built the confirm `message` as raw HTML and PrimeNG renders it via `[innerHTML]` (no `escape` option in this version), so the GitHub-controlled `environmentName` / `repositoryNameWithOwner` were unescaped. An environment named `` would run script in the reviewer's session. Now HTML-escaped before interpolation. (The decline copy was already safe — it uses `{{ }}` in the template.)

### ✅ Re-verified the 3 resolved threads — fixes are correct
- Lock + re-check in `ApprovalDecisionWriter` (first decision wins; losing writer no-ops). 
- Deferral persists PENDING rows + emails **before** stamping `DEFERRED_TO_REVIEWERS` (stamp is the idempotency gate). 
- `pending-approvals` returns `PENDING` **and** `FAILED_AT_GITHUB`. 
Authorization for User-type rules is sound (non-reviewer → 403, self-review → 403, no IDOR — authz is re-derived from the deployment's env rule, not the request id).

### Remaining findings — your call (design-level; not auto-fixed)

**[HIGH] Team-rule + GitHub rejection strands the deployment with no in-app recovery.** For a Team-type required reviewer, `reviewers.userLogins()` is empty, so the legacy impersonate-creator path runs; if GitHub rejects it, a `FAILED_AT_GITHUB` AUTO row is written but `notifyReviewers` is skipped, and the creator's in-app retry hits the `!userLogins().contains(login)` → 403 (`DeploymentReviewActionService:104`). The deployment is then only resolvable on GitHub directly. This is squarely the Phase-4 team-expansion gap — worth a follow-up issue, and maybe a "resolve on GitHub" hint in the UI for team-rule failures.

**[MEDIUM] No `UNIQUE(helios_deployment_id, reviewer_login)` + ineffective lock.** `ApprovalService` has no `@Transactional`, so `lockByDeploymentAndReviewerLogin` in `notifyReviewers` acquires/releases its `PESSIMISTIC_WRITE` lock within the SELECT (serializes nothing). With no unique constraint, a redelivered/concurrent `waiting` event can insert two rows for the same `(deployment, reviewer)`, and the later `lockByDeploymentAndReviewerLogin().Optional` then throws `NonUniqueResultException` on that reviewer's click. The find-or-create logic already assumes one row per pair — recommend adding the unique constraint in a new migration (it also makes find-or-insert race-safe at the DB level).

**[MEDIUM] Webhook auto-approve calls GitHub before taking the row lock** (acknowledged in the `ApprovalDecisionWriter` Javadoc). On NATS redelivery / GitHub webhook retry, two deliveries can both pass the unlocked `autoApprovalDecision == null` check and both POST to GitHub (second usually 422s); the lock+recheck keeps the DB authoritative but can leave orphaned PENDING rows from a late `notifyReviewers`. Fine if the JetStream consumer truly serializes per-deployment; if `maxAckPending` allows concurrent dispatch, consider a transactional claim before the GitHub call.

**[MEDIUM] V54 enables `DEPLOYMENT_APPROVAL_REQUEST` for *all* users**, including never-logged-in/synced accounts — diverges from V34's `WHERE has_logged_in = TRUE`. The comment says this is deliberate ("default-on"); flagging the divergence in case emailing never-logged-in accounts is unintended.

**[LOW]** `GitHubService.reviewPendingDeployment` omits `X-GitHub-Api-Version` and uses `Accept: application/json` while the rest of the file uses `application/vnd.github+json` + `GITHUB_API_VERSION` — pins this critical call to GitHub's rolling default. Hardening, not a bug. · **[LOW]** decline dialog's confirm button has no loading/disabled state. · **[LOW]** `profile-nav-section` `router.events` subscription isn't unsubscribed (matches existing precedent).

Migrations/entities/indexes, error propagation, and `noRollbackFor` all checked out. Happy to implement the unique-constraint migration if you want it in this PR.

PR #1182, opened 2026-07-03
refactor: explicit per-repository query filtering + @Transactional cleanup (Option B)

krusche:
Large, incremental, **test-first** refactor implementing the approved plan (`docs`: splendid-forging-pascal). It replaces the fragile Hibernate `gitRepositoryFilter` "magic" — the root cause of the 1.7.0→1.7.2 cross-repo data leak — with **explicit per-query repository filtering**, and strips `@Transactional` back to only where it is mechanically required.

Every step is its own commit and keeps the full server suite green; one commit temporarily injects a leak to prove the guards catch it (reverted in the same investigation, never committed).

## Why

The old per-tenant isolation relied on a Hibernate `@Filter` silently enabled per request. After a refactor it stopped applying under Open-Session-in-View and leaked **every repository's** data into every project view — invisible in single-repo staging. Ambient, transaction-scoped filtering is too easy to break silently. This PR makes tenant scoping explicit at each query and removes the magic entirely.

It also closes a **latent cross-repo read**: the Hibernate `@Filter` never applied to `findById`/PK loads, so `getEnvironmentById(idFromAnotherRepo)` (and the equivalents for workflows/deployments/PRs/…) leaked before this change. The new `findByIdAndRepositoryRepositoryId` finders scope those explicitly.

## What changed

**A — Test harness + guards**
- New full-context `HeliosIntegrationTest` (`@SpringBootTest` + MockMvc on zonky Postgres + real Flyway; externals disabled, GitHub clients mocked).
- Per-endpoint scoping guard ITs (Environment/Branch/Workflow): seed two repos, assert list endpoints return only the `X-REPOSITORY-ID` repo, no header → empty, cross-repo `findById` → 404. These pass on both the old (filter) and new (explicit) code — the true cross-refactor guard.

**B — Explicit per-query filtering** (one commit per service)
- `Environment`, `Branch`, `Workflow`, `WorkflowRun`, `Deployment`, `PullRequest` (incl. the paginated `Specification`), `ReleaseInfo` reads now pass the repo id from `RepositoryContext` to a scoped finder.
- Tenant reads **never** fall back to `findAll()`: no repo context → empty list (per maintainer guidance — `findAll` is too expensive and leak-prone).

**C — Fix write-on-GET**
- `GET /settings` no longer INSERTs a settings row. New read-only `getGitRepoSettingsByRepositoryId` (returns transient defaults, no `save`); `getOrCreate` retained for the write callers.

**D — Remove the Hibernate filter infra**
- Dropped `@FilterDef`/`@Filter` from all entities + `package-info.java`, deleted `RepositoryFilterTransactionConfig` and the filter-mechanism test. `RepositoryInterceptor` now only sets/clears `RepositoryContext` (and clears on no-header, defending against stale ThreadLocals on reused worker threads). OSIV kept on (disabling it is a separate, larger effort).

**E — `@Transactional` cleanup**
- A transaction boundary is not free: it holds a DB connection (and any row locks) for the method's whole duration — harmful when the method spans a GitHub/NATS network call — and its commit-time dirty-check flushes every modified managed entity, risking unintended writes. Since writes almost always succeed, "atomicity just in case" rarely pays for that.
- Removed the class-level `@Transactional` from `GitHubService` (a pure API gateway with zero JPA — it was holding a DB tx across GitHub network calls) and from all read-heavy services. Every write already calls `save()`/`saveAndFlush()`/`delete(entity)` explicitly (Spring Data's per-call transaction persists each), so removal changes nothing about *whether* data is written.
- `@Transactional` is kept **only where Spring throws without it** — the two derived delete queries (`BranchService.deleteBranchByNameAndRepositoryId`, `ReleaseInfoService.deleteReleaseCandidateByName`).

**F — Delete dead code**
- Removed the now-unused unscoped finders (`findByName`, `findAllByState`, `findAllByOrderBy…`, the no-repo `TestCaseStatistics` finder) and the empty `IssueRepository` (verified zero callers across main + test).

## Verification

- Full `:application-server` server suite green.
- **Guards proven effective:** temporarily reverting one migrated query to an unscoped `findAll()` made `EnvironmentScopingIT` fail (2/5); restoring the scoped finder returned it to green.

## Behavior preserved / flagged (not changed here)
- `RepositoryService.getAllRepositories` stays intentionally cross-repo (repo picker).
- `User` lookups are globally unique (`unique_user_login`), so global lookups after filter removal are safe.
- OSIV remains enabled — disabling it is a separate follow-up.

## Remaining (owner)
- Merge → staging deploy → confirm multi-repo scoping on the running app (staging has ≥2 repos: `GET /api/environments/enabled` with `X-REPOSITORY-ID` returns only that repo's rows) and no `gitRepositoryFilter` errors.


krusche:
@Claudia-Anthropica review

Claudia-Anthropica:
@krusche acknowledged, queuing this up for immediate review!

Claudia-Anthropica (changes requested):
@krusche I couldn't place the inline comments because GitHub rejected one of the diff positions, so I'm leaving the findings here. The query migration is going in the right direction, but workflow write/config paths still accept global workflow IDs after the inherited filter was removed: `EnvironmentService.updateEnvironment` can attach a deployment workflow from another repo, `WorkflowService.updateWorkflowLabel` can relabel a workflow outside the selected repo, and `WorkflowGroupService.updateWorkflowGroups` can add foreign workflows through `findAllById`/`findById`. Please scope those workflow lookups to the current/requested repository and add regression tests; Codacy Static Code Analysis is also still red, but I did not duplicate the existing reviewdog comments.

krusche:
All three findings from the review are addressed, each with a regression IT that asserts a cross-repository id is rejected (no mutation) while the same-repository path still succeeds:

1. `EnvironmentService.updateEnvironment` — the deployment-workflow attach now resolves via `findByIdAndRepositoryRepositoryId` against the environment's own repository (55335da9). Guard: `EnvironmentUpdateWorkflowScopingIT`.
2. `WorkflowService.updateWorkflowLabel` — routed through the repository-scoped `findScopedById` (75ac3e05). Guard: `WorkflowLabelScopingIT`.
3. `WorkflowGroupService.updateWorkflowGroups` — membership workflow lookups scoped via a new `findByIdInAndRepositoryRepositoryId` plus `findByIdAndRepositoryRepositoryId` (a4c0e0aa). Guard: `WorkflowGroupScopingIT`.

Full server suite green (470 tests, 0 failures). Codacy Static Analysis: server test sources are excluded from its default Checkstyle, which rejected the standard `*IT` integration-test suffix and duplicated the repo's own `server/checkstyle.xml`.

PR #1197, opened 2026-07-15
feat(cleanup): 90-day hard retention cap + default-branch exemption for keep-N pruning

krusche:
## Motivation

The nightly cleanup deletes rows, but the keep-N policies retain up to ~6 runs per ⟨repo, workflow, branch⟩ **forever** as long as the branch exists. On prod this meant 473 live branches holding 15.3M `test_case` rows — 47% of them from runs older than 30 days, the oldest from February 2025.

## Changes

- **Default-branch exemption**: runs on a repository's default branch (`develop`/`main`) are no longer touched by keep-N pruning — full history is kept there.
- **Hard retention cap**: new `cleanup.workflow-run.max-age-days` (default **90**, overridable via `CLEANUP_WORKFLOW_RUN_MAX_AGE_DAYS`) deletes every run older than the cap, regardless of status, branch (default branches included) or keep window. Runs after the keep-N policies in the same 01:00 job and honours the existing dry-run flag.

Net retention semantics: feature branches keep their newest 2 runs per status bucket for up to 90 days; default branches keep everything for 90 days; nothing survives past 90 days.

## Notes

- `deployment.workflow_run_id` / `helios_deployment.workflow_run_id` have no FK to `workflow_run` (V42 dropped it), so the cap's cascade only touches `test_suite`/`test_case`/`workflow_job`/`workflow_run_pull_requests` — same as the existing keep-N purge.
- Space freed by these deletes is reused in-place by Postgres; returning it to disk still needs a periodic reindex/repack (done manually on prod today, 28 GB → 12 GB).

## Testing

- Unit tests for the task wiring (delete mode, dry-run, unset, invalid config).
- New zonky embedded-Postgres integration test for the SQL semantics: default-branch exemption join, keep-window pruning, cap deletion incl. cascade to `test_suite`/`test_case`.


server/application-server/src/main/java/de/tum/cit/aet/helios/workflow/WorkflowRunRepository.java
```diff
@@ -299,6 +307,31 @@ int purgeObsoleteRuns(@Param("keep") int keepPerCombo,
...
+   * Hard retention cap: deletes every workflow run older than
+   * {@code maxAgeDays}, regardless of status, branch (default branches
+   * included) or any keep-N policy. Child rows in test_suite → test_case,
+   * workflow_job and workflow_run_pull_requests cascade via their FKs.
+   */
+  @Modifying
+  @Transactional
+  @Query(value = """
+      DELETE FROM workflow_run
+      WHERE created_at < now() - (:maxAgeDays * interval '1 day')
```
Claudia-Anthropica: @krusche [high] This executes the entire age backlog as one `DELETE` inside `purge()`'s outer transaction, so the first 90-day rollout can cascade through millions of `test_case` rows while holding locks and WAL until one commit. Please delete a bounded number of run IDs per transaction; the orphan cleanup already documents why batching is necessary for exactly this data shape.

In `server/application-server/src/main/java/de/tum/cit/aet/helios/workflow/WorkflowRunRepository.java`, `purgeRunsOlderThan` deletes the complete max-age backlog in one transaction and can cascade through millions of child rows. Replace it with indexed, bounded batch selection/deletion and run each batch in a separate transaction, including restructuring the outer transaction in `WorkflowRunCleanupTask.purge()`, then add a batch-boundary test.

</details>

server/application-server/src/main/java/de/tum/cit/aet/helios/workflow/WorkflowRunRepository.java
```diff
@@ -198,17 +198,19 @@ List<WorkflowRun> findLatestForDeploymentReadiness(
...
-                     ORDER BY created_at DESC
+                     PARTITION BY wr.repository_id, wr.workflow_id, wr.head_branch
+                     ORDER BY wr.created_at DESC
                  ) AS rn
-          FROM workflow_run
-          WHERE test_processing_status IS NOT DISTINCT FROM :tps
+          FROM workflow_run wr
+          LEFT JOIN repository gr ON gr.repository_id = wr.repository_id
+          WHERE wr.test_processing_status IS NOT DISTINCT FROM :tps
+            AND wr.head_branch IS DISTINCT FROM gr.default_branch
```
Claudia-Anthropica: @krusche [medium] Moving the default-branch predicate into `ranked` makes `previewSurvivorRunIds()` omit every exempt default-branch run, even though its contract and the dry-run log say these IDs survived. Keep exempt rows in the survivor result while excluding them only from `deletable`, and cover that result in the integration test.

In `server/application-server/src/main/java/de/tum/cit/aet/helios/workflow/WorkflowRunRepository.java`, `previewSurvivorRunIds` filters default-branch runs out before computing survivors, so dry-run output omits runs that are explicitly exempt. Keep default-branch rows in the returned survivor set while applying the exemption only to deletion eligibility, and add an integration assertion for the default branch.

</details>

Claudia-Anthropica (changes requested):
@krusche The retention semantics look right, but the production deletion path needs batching before it is safe to roll out. The dry-run survivor preview also needs to include exempt default-branch runs; details inline.

krusche:
Deep-review pass (multi-agent adversarial review) surfaced several issues, all fixed in a9af2df7:

1. **NULL-handling regression in the exemption** — `head_branch IS DISTINCT FROM default_branch` silently exempted NULL-branch runs in repos with a NULL `default_branch` (and runs whose repository row is gone) from all keep-N pruning. Now `(head_branch = default_branch) IS TRUE`: only a real name match exempts.
2. **Dangling deployment links** — the cap deleted runs still referenced by `helios_deployment`/`deployment` (FK-less since V41/V42), breaking the deployment → build trail the orphan sweep deliberately preserves. The cap (and its dry-run count) now excludes deployment-referenced runs.
3. **Survivor preview omitted default-branch runs** — the exemption sat in the ranked CTE, so DRY-RUN 'rows will survive' silently dropped exempt runs. Moved into the deletable filter; survivors + deletable partition the run set again.
4. **Unbatched backlog delete** — first non-dry run would have deleted the whole >90d backlog (~5–6M cascading rows on prod) in one transaction inside `@Transactional purge()`. The cap now deletes in 5000-run batches, each its own transaction, and `purge()` no longer wraps everything in one tx.
5. **No disable switch** — javadoc claimed `null` disables, but the yml default made null unreachable and `0` warned 'invalid' daily. `CLEANUP_WORKFLOW_RUN_MAX_AGE_DAYS=0` now cleanly disables; negatives warn; docs aligned.
6. **Dry-run totals** — the cap now feeds its would-be-deleted count into the finishing total like the keep-N previews do (overlap with policy previews is called out in the log line).

Known accepted limitation (documented in the props javadoc): fork-PR runs whose head branch shares the default branch's name are exempt from keep-N too — the schema stores no head repository to tell them apart; the 90-day cap bounds them.

Note kept as-is by design: the cap ships **enabled at 90 days by default** — that is the point of this PR for our deployment; other operators can set `CLEANUP_WORKFLOW_RUN_MAX_AGE_DAYS=0`, and the existing dry-run flag gates it.

Verified: 508 application-server tests green, incl. extended zonky integration tests (NULL default-branch pruning, survivor-preview completeness, deployment exclusion, batch bounding); checkstyle 10.21.0 clean.

server/application-server/src/main/java/de/tum/cit/aet/helios/workflow/WorkflowRunRepository.java
Claudia-Anthropica: @krusche Looks good now, thanks. The exemption is applied only to deletion eligibility and the survivor preview is covered.

server/application-server/src/main/java/de/tum/cit/aet/helios/workflow/WorkflowRunRepository.java
Claudia-Anthropica: @krusche Nice fix — the cap now deletes in bounded, separately committed batches.

Claudia-Anthropica (approved):
@krusche All feedback addressed, nice work.

PR #1198, opened 2026-07-19
fix(deployment): surface GitHub approval failures instead of a generic 500

krusche:
## Motivation

Clicking **Approve**/**Decline** on a pending deployment showed a red toast reading *"Error: An internal server error occurred"* — with no hint of what actually went wrong. On prod the real cause was a GitHub **HTTP 401** (the deployer's Keycloak-brokered GitHub token had expired), which the service correctly turns into a `502 "GitHub rejected the request: … HTTP 401"`.

That message never reached the user: `GlobalExceptionHandler`'s generic `@ExceptionHandler({Exception.class, IOException.class})` catches `ResponseStatusException` too (advice handlers run before Spring's `ResponseStatusExceptionResolver`), so **every** `ResponseStatusException` in the app was being flattened to a flat `500 "internal server error"`, discarding both the intended status and the message.

Prod log for the reported failure:

```
GitHubService: GitHub pending-deployment approved failed for run …: HTTP 401
DeploymentReviewActionService: In-app approve by @krusche for deployment 9436 failed at GitHub: HTTP 401
GlobalExceptionHandler: An internal server error occurred
  → ResponseStatusException: 502 BAD_GATEWAY "GitHub rejected the request: … HTTP 401"
```

## Changes

- **`GlobalExceptionHandler`**: add a dedicated `@ExceptionHandler(ResponseStatusException.class)` that preserves the intended status + reason. These are deliberate, handled outcomes, so they're kept out of Sentry (5xx logged at WARN, 4xx at DEBUG). This fixes the misleading error for **all** endpoints that throw `ResponseStatusException`, not just approvals.
- **`GitHubReviewException`** (new, extends `IOException`): carries GitHub's HTTP status. The review path uses it to detect a **401** and return an actionable reason — *"your GitHub authorization has expired. Please sign out of Helios and sign in again, then retry."* — while still returning **502** to the client (returning 401 would make the client treat the user's Helios session as expired and bounce them to login).

## Scope / follow-up

This makes the failure **legible**; it does **not** stop the 401s. The underlying cause — Keycloak stores the GitHub token at login and never refreshes it, and GitHub App user tokens expire after 8h, so auto- and in-app approval both fail once the token is stale — is being addressed separately (Keycloak refresh-token integration).

## Testing

- `DeploymentApprovalControllerErrorHandlingTest` (new, `@WebMvcTest`): a service-thrown `502` reaches the client as `502` with its reason intact; an unexpected exception still falls back to `500`.
- `DeploymentReviewActionServiceTest` (new case): a GitHub `401` yields `502` with the actionable "expired / sign in again" message and still records a `FAILED_AT_GITHUB` audit row.
- Existing `GitHubServiceTest` failure case unchanged (`GitHubReviewException` is an `IOException` with the same message).

### DIFF
--- a/server/application-server/src/main/java/de/tum/cit/aet/helios/deployment/approval/DeploymentReviewActionService.java
+++ b/server/application-server/src/main/java/de/tum/cit/aet/helios/deployment/approval/DeploymentReviewActionService.java
@@ -3,6 +3,7 @@ package de.tum.cit.aet.helios.deployment.approval;
 import de.tum.cit.aet.helios.environment.Environment;
 import de.tum.cit.aet.helios.environment.EnvironmentService;
 import de.tum.cit.aet.helios.environment.EnvironmentService.ReviewerResolution;
+import de.tum.cit.aet.helios.github.GitHubReviewException;
 import de.tum.cit.aet.helios.github.GitHubService;
 import de.tum.cit.aet.helios.heliosdeployment.HeliosDeployment;
 import de.tum.cit.aet.helios.heliosdeployment.HeliosDeploymentRepository;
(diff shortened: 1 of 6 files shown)


Claudia-Anthropica (approved):
@krusche This cleanly preserves intentional response statuses and gives expired GitHub authorization failures an actionable message while keeping the client-facing status at 502. Nice focused tests around both the controller response and audit-row behavior; approving as-is.

krusche:
we should still work on this and get it fixed: not stale

PR #1199, opened 2026-07-19
feat(auth): self-refreshing GitHub user tokens for deployment approvals

krusche:
## Motivation

Deployment approvals (both Helios' automatic approval and the in-app Approve/Decline buttons) act on GitHub using a GitHub **user** token obtained via Keycloak token-exchange. GitHub App user tokens expire after **8 hours**, and Keycloak never refreshes brokered tokens — so a few hours after the reviewer last logged in, GitHub returns `HTTP 401` and approvals fail (the "internal server error" reported on prod). It can never be "always automatic" as built, because the token only refreshes on interactive login.

This PR makes Helios **own the refresh loop** so a valid user token is always available, independent of when the user last logged in.

## Approach (Helios-side refresh)

Keycloak config-only refresh is not possible for the `github` provider (confirmed against Keycloak source + issues), so:

1. **Seed** each user's GitHub *refresh token* once from Keycloak's broker retrieve-token endpoint (`GET /broker/github/token`), reached headlessly via an impersonation token-exchange.
2. **Refresh** directly against GitHub (`grant_type=refresh_token`) using the App's client id/secret, caching the ~8h access token and **persisting the rotated refresh token** each time (GitHub rotates them).
3. On an unrecoverable refresh (revoked/expired) surface `GitHubReauthRequiredException` so the reviewer is told to sign in again.

Refresh tokens are 6-month credentials, so they're **encrypted at rest** (AES-GCM).

## What's here (components)

- `TokenCipher` — AES-GCM encrypt/decrypt; missing key degrades only token ops (never crashes boot, never stores plaintext).
- `github_user_token` table (**V59**) + entity/repo — one encrypted row per GitHub login.
- `GitHubOAuthTokenClient` — GitHub refresh grant (rotation-aware; distinguishes re-auth vs transient).
- `KeycloakBrokerTokenClient` — seeds the refresh token via retrieve-token (parses JSON or form-encoded).
- `GitHubUserTokenService` — cache → refresh → seed orchestration.
- `GitHubService` rewired: approvals + release drafts now use `getValidAccessToken(...)`.

## ⚠️ Required before merge/deploy (ops)

- **`GITHUB_CLIENT_SECRET`** — the login GitHub App's OAuth client secret (App must have "Expire user authorization tokens" **on** so refresh tokens are issued).
- **`HELIOS_TOKEN_ENCRYPTION_KEY`** — `openssl rand -base64 32`. *(Missing key no longer crashes the app — it logs a warning and approvals fail until set — but it must be set for the feature to work.)*
- **Keycloak: grant the `helios-token-exchange` client the retrieve-token permission on the `github` IdP** (Identity Providers → github → Permissions → the `token` permission → add its policy). **Confirmed needed** — a read-only spike currently returns `403 "Client [helios-token-exchange] not authorized to retrieve tokens from identity provider [github]"`. (Headless impersonation token-exchange itself already returns 200.)

## Relationship to #1198

#1198 (error surfacing) and this PR both touch `GitHubService`. The final **actionable "GitHub authorization expired — sign in again" message** for the re-auth case (mapping `GitHubReauthRequiredException`) lands as a small follow-up once #1198 merges — tracked in the plan (`docs/superpowers/plans/2026-07-19-github-user-token-refresh.md`, Task 9).

## Testing

Unit tests for every new component (cipher round-trip; refresh parsing + rotation + re-auth vs transient; seed parsing for JSON *and* form-encoded bodies + the 403 path; service cache/refresh/seed/re-auth branches; rewired `GitHubService`). New embedded-Postgres integration test for the V59 store. Full `:application-server:test` suite green locally.

### DIFF
--- a/docs/superpowers/plans/2026-07-19-github-user-token-refresh.md
+++ b/docs/superpowers/plans/2026-07-19-github-user-token-refresh.md
@@ -1,0 +1,292 @@
+# GitHub User-Token Refresh Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Make Helios always hold a valid GitHub *user* access token for deployment approvals, by owning the refresh loop itself — so auto- and in-app approval keep working regardless of how long ago the reviewer logged in.
+
+**Architecture:** Helios stops depending on Keycloak's (non-refreshing) token-exchange for the live GitHub token. Instead it persists each user's GitHub **refresh token** (seeded once from Keycloak's broker *retrieve-token* endpoint), then refreshes directly against GitHub's OAuth token endpoint on demand, caching the ~8 h access token and rotating the refresh token on every use. Refresh tokens are 6-month credentials, so they are encrypted at rest.
+
+**Tech Stack:** Java 21+, Spring Boot, Spring Data JPA, Flyway, OkHttp, Jackson, JUnit 5 + Mockito, zonky embedded Postgres (integration tests), Keycloak 26.1.3, GitHub App OAuth.
+
+## Global Constraints
+
+- Java source compiled on JDK 25; language level per existing `build.gradle` — match surrounding code.
+- Checkstyle (`server/checkstyle.xml`, Google-based): **max line length 100**, 2-space indent, alphabetical imports (static first), Javadoc on public types.
+- Dependencies pinned exactly — **never** use `^`/version ranges; reuse libraries already on the classpath (OkHttp, Jackson) — add none.
+- Flyway migrations are immutable and sequential; next free version is **V59**. Never edit an applied migration.
+- Deployment-referenced integrity: do not weaken existing approval-path behaviour (audit rows, `FAILED_AT_GITHUB`, self-review checks).
+- All new secrets are provided via environment variables (compose), never committed.
+- Every task ends green: `server/gradlew -p server :application-server:test` (targeted with `--tests`).
+
+---
+
+## Prerequisites (ops / prod config — NOT code; do before Task 2)
+
+These are operator steps. Document exact values in the PR description; do not hard-code secrets.
+
+**P1. GitHub App OAuth client secret.** The refresh grant needs the login GitHub App's `client_id` + `client_secret`. Helios already has `GITHUB_CLIENT_ID`; provision the matching secret as `GITHUB_CLIENT_SECRET`. Confirm the App has **"Expire user authorization tokens" enabled** (Developer settings → the App → Optional features) — that is what makes GitHub issue refresh tokens at all.
+
+**P2. Token-encryption key.** Provision `HELIOS_TOKEN_ENCRYPTION_KEY` = base64-encoded 256-bit key (`openssl rand -base64 32`). App fails fast at startup if unset (see Task 3) so refresh tokens are never written in plaintext.
+
+**P3. Keycloak retrieve-token permission — CONFIRMED by the Task-1 spike (2026-07-19).** In realm `helios`, the `github` IdP already has `Store tokens = ON`, and a headless impersonation token-exchange for a user already succeeds (spike Step A → HTTP 200). The only missing grant: **Identity Providers → github → Permissions → enable, then the `token` (retrieve-token) permission → add a client policy for `helios-token-exchange`.** Without it the retrieve endpoint returns exactly:
+
+```
+StepB(retrieve-token /broker/github/token) http=403
+{"errorMessage":"Client [helios-token-exchange] not authorized to retrieve tokens from identity provider [github]."}
+```
+
+This is a mirror of the impersonation permission that already lets Step A work; low-risk and reversible. After granting it, re-run the spike (`scratchpad/kc_spike.sh`) — Step B must return 200 with a body containing `refresh_token` — before implementing Task 5.
+
+---
+
+## Risk & the Task-1 spike
+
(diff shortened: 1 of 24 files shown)


server/application-server/src/main/java/de/tum/cit/aet/helios/auth/github/token/GitHubUserTokenService.java
```diff
@@ -0,0 +1,93 @@
...
+
+  private final GitHubUserTokenRepository repository;
+  private final TokenCipher tokenCipher;
+  private final KeycloakBrokerTokenClient brokerClient;
+  private final GitHubOAuthTokenClient oauthClient;
+
+  @Transactional
+  public String getValidAccessToken(String githubLogin) throws IOException {
+    OffsetDateTime now = OffsetDateTime.now();
+    Optional<GitHubUserToken> existing = repository.findByGithubLogin(githubLogin);
```
Claudia-Anthropica, line 45: @krusche [high] This read does not lock or otherwise serialize refreshes for a login. GitHub documents refresh tokens as single-use, so two deployments for the same reviewer can both read the same token; one rotates it and the other then fails, while concurrent first-use calls can also race into the unique constraint. Serialize the full lookup/refresh/persist sequence per normalized login across application instances, including the no-row case.

In `server/application-server/src/main/java/de/tum/cit/aet/helios/auth/github/token/GitHubUserTokenService.java`, concurrent calls can consume the same single-use GitHub refresh token or race while creating the first row. Add database-backed per-login serialization around lookup, refresh, and persistence, handle the absent-row case, and add a concurrency test proving only one refresh grant runs.

</details>

server/application-server/src/main/java/de/tum/cit/aet/helios/auth/github/token/GitHubUserTokenService.java
```diff
@@ -0,0 +1,93 @@
...
+          && row.getAccessTokenExpiresAt() != null
+          && row.getAccessTokenExpiresAt().isAfter(now.plus(EXPIRY_MARGIN))) {
+        return tokenCipher.decrypt(row.getAccessTokenEnc());
+      }
+
+      if (row.getRefreshTokenEnc() != null
+          && (row.getRefreshTokenExpiresAt() == null
+              || row.getRefreshTokenExpiresAt().isAfter(now))) {
+        GitHubUserTokenRecord refreshed =
+            oauthClient.refresh(tokenCipher.decrypt(row.getRefreshTokenEnc()));
```
Claudia-Anthropica, line 60: @krusche [high] A rejected stored refresh token is propagated immediately, so signing in again cannot actually recover: Keycloak gets the new token, but every later call keeps retrying the stale encrypted database token. On `bad_refresh_token`, reseed from Keycloak and replace the row; only surface reauth if that fresh seed is also unavailable or rejected.

In `server/application-server/src/main/java/de/tum/cit/aet/helios/auth/github/token/GitHubUserTokenService.java`, a rejected persisted refresh token permanently wins over a newer token stored by Keycloak after login. Catch the token-specific reauthorization exception from the stored-token refresh, reseed and persist through `KeycloakBrokerTokenClient`, and add a test that a post-login seed repairs the stale row.

</details>

server/application-server/src/main/java/de/tum/cit/aet/helios/auth/github/token/GitHubOAuthTokenClient.java
```diff
@@ -0,0 +1,97 @@
...
+
+    try (Response response = okHttpClient.newCall(request).execute()) {
+      ResponseBody responseBody = response.body();
+      String content = responseBody == null ? "" : responseBody.string();
+      if (!response.isSuccessful()) {
+        throw new IOException("GitHub token refresh failed: HTTP " + response.code());
+      }
+
+      JsonNode json = objectMapper.readTree(content);
+      if (json.hasNonNull("error")) {
```
Claudia-Anthropica, line 76: @krusche [medium] Every OAuth error is classified as user reauthorization, including app-side failures such as `incorrect_client_credentials` or an unsupported grant. Signing out cannot fix those, and this classification would also make the service's reseed path misleading. Map only refresh-token rejection codes to `GitHubReauthRequiredException`; keep configuration and malformed/transient responses as ordinary `IOException`s.

In `server/application-server/src/main/java/de/tum/cit/aet/helios/auth/github/token/GitHubOAuthTokenClient.java`, all OAuth error responses are incorrectly treated as requiring user login. Whitelist only refresh-token rejection errors for `GitHubReauthRequiredException`, map configuration and protocol errors to `IOException`, and add tests for both categories.

</details>

docs/superpowers/plans/2026-07-19-github-user-token-refresh.md
```diff
@@ -0,0 +1,292 @@
+# GitHub User-Token Refresh Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
```
Claudia-Anthropica, line 3: @krusche [medium] This line is an instruction aimed directly at automated agents, including a required tool/workflow choice. Repository documentation should describe the implementation for readers without attempting to control an agent's execution environment; please remove this directive or rewrite it as neutral plan context.

In `docs/superpowers/plans/2026-07-19-github-user-token-refresh.md`, line 3 directs automated agents to use specific skills and workflows. Remove the agent-directed directive and leave only neutral, human-readable implementation-plan context.

</details>

Claudia-Anthropica (changes requested):
@krusche The encrypted persistence and component-level tests are a solid base, but the rotation path needs per-user serialization and a recovery path after reauthentication before this is safe to merge. OAuth configuration failures also need to remain distinct from user reauthorization, and the agent-directed instruction in the plan should be removed. Codacy is currently failing as well.

krusche:
we should still work on this and get it fixed: not stale

